PR #3, opened 2023-05-07
Test also a pool where DAI is second

oldchili:


src/tests/FlapperUniV2.t.sol
```diff
@@ -83,22 +88,26 @@ contract FlapperUniV2Test is DssTest {
...
     address     immutable DAI           = ChainlogLike(LOG).getAddress("MCD_DAI");
     address     immutable MKR           = ChainlogLike(LOG).getAddress("MCD_GOV");
+    address     immutable USDC          = ChainlogLike(LOG).getAddress("USDC");
+    address     immutable LINK          = ChainlogLike(LOG).getAddress("LINK");
     address     immutable PAUSE_PROXY   = ChainlogLike(LOG).getAddress("MCD_PAUSE_PROXY");
     VatLike     immutable vat           = VatLike(ChainlogLike(LOG).getAddress("MCD_VAT"));
     VowLike     immutable vow           = VowLike(ChainlogLike(LOG).getAddress("MCD_VOW"));
     EndLike     immutable end           = EndLike(ChainlogLike(LOG).getAddress("MCD_END"));
     SpotterLike immutable spotter       = SpotterLike(ChainlogLike(LOG).getAddress("MCD_SPOT"));
+    PipLike     immutable linkPip       = PipLike(ChainlogLike(LOG).getAddress("PIP_LINK"));
```
sunbreak1211, line 106: We might want to use the medianizer instead of the osm.

telome: Or to keep the tests symmetrical and easier to read I think it would be fine to use another mock medianizer for LINK.

oldchili: Alright, changed to using mock medainizer also for LINK.
It is also getting offboarded now so relying on the oracle is not a good idea.

telome (approved):

sunbreak1211 (approved):
Looks good

PR #4, opened 2023-05-09
CS Audit

oldchili:


README.md
```diff
@@ -0,0 +1,29 @@
...
+
+### OracleWrapper
+
+Allows for scaling down an oracle price by a certain value. This can be useful when the `gem` is a redenominated version of an existing token, which already has a reliable oracle.
+
+### Notes:
+
+* The swapped amount (`lot`, or `vow.bump` in case the `vow` is the `msg.sender`) is received in 10^45 (`RAD`) resolution, and must be a multiple of 10^27 (`RAY`).
+
+* Availability and accounting of the drawn `DAI` is the responsibility of the `msg.sender`. In case that is the `vow`, at the time of a `kick`, it is expected to only hold the swapped amount (`lot`, or `vow.bump`) over the configured flapping threshold (`vow.hump`). As a `kick` operation also draws `DAI` for depositing in the pool (and not only for swapping), it can in practice reduce the Surplus Buffer to below that threshold.
```
sunbreak1211, line 26:
```suggestion
* Availability and accounting of the withdrawn `DAI` is the responsibility of the `msg.sender`. In case that is the `vow`, at the time of a `kick`, it is expected to only hold the swapped amount (`lot`, or `vow.bump`) over the configured flapping threshold (`vow.hump`). As a `kick` operation also withdraws `DAI` for depositing in the pool (and not only for swapping), it can in practice reduce the Surplus Buffer to below that threshold.
```

sunbreak1211:
One general thing to comment is that the `Vow` is mentioned like something that could be or not the caller. I'd say we chose the "Flapper" path to be used via the `Vow`. We even had to make the `kick` method conformant with the `Vow` taking an unused parameter. So my suggestion would be taking the documentation more in the direction that this has been built only to be used by the existing `Vow` and be more specific about the caveats.

README.md
telome:
```
it is expected to only hold the swapped amount (`lot`, or `vow.bump`) over the configured flapping threshold (`vow.hump`).
```
Not entirely sure what this is trying to say. I would maybe rephrase it so say that the `msg.sender` is expected to hold at least 2.2*`lot` for a successful execution of `kick()`, given the additional 1-1.2*`lot` withdrawn for the deposit. And in the case where the `msg.sender` is the `vow`, this additional withdrawal implies that the Surplus Buffer can end up being reduced to below the `vow.hump` threshold after `kick()`.

oldchili: I broke that part to two, Let me know if it is easier to understand.
Would prefer not to explicitly mention the 2.2 value of the sanity check, as it doesn't come from some well defined restriction. Also, whould rather not need to change here if that value changes.
My main idea was to emphasize that `vow.hump` threshold is broken deliberately and prevent queries, don't think the amount itself changes so much.

oldchili:
> So my suggestion would be taking the documentation more in the direction that this has been built only to be used by the existing `Vow` and be more specific about the caveats.

Done, now sender is as assumed as the `vow`.

README.md
```diff
@@ -0,0 +1,31 @@
+# Dss Flappers
+
+Implementantions of MakerDao surplus auctions, triggered on `vow.flap`. The current featured Flapper is `FlapperUniV2`.
```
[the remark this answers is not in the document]
sunbreak1211: I like "Implementantions" though :)

sunbreak1211:
Looks better, but IMO the `drawn` term is a bit confusing. I'd change it to "withdrawn" as it gives more a sense of taking out from an existing surplus which in this case is totally correct wording as the `vow` won't give anything to the Flapper that is not real surplus. This is different than using `suck` where you can't actually ensure you might be taking bad debt.

oldchili:
> Looks better, but IMO the `drawn` term is a bit confusing. I'd change it to "withdrawn" as it gives more a sense of taking out from an existing surplus which in this case is totally correct wording as the `vow` won't give anything to the Flapper that is not real surplus. This is different than using `suck` where you can't actually ensure you might be taking bad debt.

Ah, I somehow missed it.
Done.

src/FlapperUniV2.sol
```diff
@@ -133,14 +135,33 @@ contract FlapperUniV2 {
...
+    function _getReserves() internal returns (uint256 reserveDai, uint256 reserveGem) {
         (uint256 _reserveA, uint256 _reserveB,) = pair.getReserves();
         (reserveDai, reserveGem) = daiFirst ? (_reserveA, _reserveB) : (_reserveB, _reserveA);
+
+        uint256 _daiBalance = GemLike(dai).balanceOf(address(pair));
+        uint256 _gemBalance = GemLike(gem).balanceOf(address(pair));
+        if (_daiBalance > reserveDai || _gemBalance > reserveGem) {
+            pair.sync();
+            reserveDai = _daiBalance;
+            reserveGem = _gemBalance;
```
telome, line 147:
```suggestion
            (reserveDai, reserveGem) = (_daiBalance, _gemBalance);
```

telome (approved):
LGTM

sunbreak1211 (approved):
Looks good for merging

PR #6, opened 2023-05-15
Add deploy scripts

oldchili:


src/deploy/FlapperDeploy.sol
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    function deployOracleWrapper(
+        address pip,
+        address flapper,
+        uint256 divisor
+    ) internal returns (address wrapper) {
+        wrapper = address(new OracleWrapper(pip, flapper, divisor));
+    }
+}
```
sunbreak1211, line 56: Missing breakline

sunbreak1211, line 56:
```suggestion
}

```

src/deploy/FlapperInit.sol
```diff
@@ -0,0 +1,94 @@
...
+        FlapperUniV2Like flapper = FlapperUniV2Like(flapperInstance.flapper);
+        FlapperMomLike   mom     = FlapperMomLike(flapperInstance.mom);
+
+        // Sanity checks
+        require(flapper.vat()     == address(dss.vat),     "Flapper vat mismatch");
+        require(flapper.daiJoin() == address(dss.daiJoin), "Flapper daiJoin mismatch");
+        require(flapper.spotter() == address(dss.spotter), "Flapper spotter mismatch");
+
+        require(cfg.hop >= 5 minutes, "hop too low");
+        require(cfg.want >= WAD * 90 / 100, "want too low");
```
sunbreak1211, line 67: Maybe:
```suggestion
        require(cfg.want >= WAD * 90 / 100 && cfg.want <= WAD, "want out of range");
```

src/deploy/FlapperInit.sol
```diff
@@ -0,0 +1,94 @@
...
+
+        flapper.file("hop",  cfg.hop);
+        flapper.file("want", cfg.want);
+        flapper.file("pip",  cfg.pip);
+        flapper.rely(address(dss.vow));
+        flapper.rely(address(mom));
+
+        dss.vow.file("flapper", address(flapper));
+        dss.vow.file("bump",    cfg.bump);
```
sunbreak1211, line 76: We could probably add the hump as well. It is a value exclusive of the `flap()` execution and in fact we will need to modify it when the flapper is on production.

sunbreak1211: Nvm, that happens due to commenting line by line without the full prev read.

sunbreak1211: Well I might still feel that is better to keep in the same init function and not separated.

oldchili: what about changing the flapper? I can imagine that happening.

src/tests/FlapperMom.t.sol
```diff
@@ -49,71 +54,80 @@ contract FlapperMomTest is Test {
...
+            daiJoin:  DAI_JOIN,
+            spotter:  SPOT,
+            gem:      MKR,
+            pair:     UNIV2_DAI_MKR_PAIR,
+            receiver: PAUSE_PROXY
+        });
+        flapper = FlapperUniV2(flapperInstance.flapper);
+        mom = FlapperMom(flapperInstance.mom);
+
+        FlapperUniV2Config memory cfg = FlapperUniV2Config(5 minutes, 1e18, address(0), 0); // use random values
```
sunbreak1211, line 69: Nit: maybe better to use the full way to pass the struct so we can see the name of the params.

src/tests/FlapperMom.t.sol
```diff
@@ -49,71 +54,80 @@ contract FlapperMomTest is Test {
...
 
     function testSetOwner() public {
         vm.expectEmit(true, false, false, false);
         emit SetOwner(address(123));
-        mom.setOwner(address(123));
+        vm.prank(PAUSE_PROXY); mom.setOwner(address(123));
         assertEq(mom.owner(), address(123));
     }
 
     function testSetOwnerNonAuthed() public {
```
sunbreak1211, line 93:
```suggestion
    function testSetOwnerNotAuthed() public {
```
not related to this PR but I guess we can change.

src/tests/FlapperMom.t.sol
```diff
@@ -49,71 +54,80 @@ contract FlapperMomTest is Test {
...
 
     function testSetAuthority() public {
         vm.expectEmit(true, false, false, false);
         emit SetAuthority(address(123));
-        mom.setAuthority(address(123));
+        vm.prank(PAUSE_PROXY); mom.setAuthority(address(123));
         assertEq(mom.authority(), address(123));
     }
 
     function testSetAuthorityNonAuthed() public {
```
sunbreak1211, line 105:
```suggestion
    function testSetAuthorityNotAuthed() public {
```

src/tests/FlapperUniV2.t.sol
```diff
@@ -253,7 +274,7 @@ contract FlapperUniV2Test is DssTest {
     }
 
     function testAuthModifiers() public virtual {
-        WardsLike(address(flapper)).deny(address(this));
+        vm.prank(PAUSE_PROXY); WardsLike(address(flapper)).deny(address(this));
```
sunbreak1211, line 277: I don't think we need this anymore, as now the test is not relied.

src/deploy/FlapperInit.sol
sunbreak1211: Actually we might be fine just having an off-boarding function. So when we want to do a replacement we can off-board the actual and then do the onboarding of the new one. As the new Flapper might have other settings seems a replacement function might not be possible to generalize.

src/tests/OracleWrapper.t.sol
```diff
@@ -17,52 +17,65 @@
...
 
     function setUp() public {
-        medianizer = new MockMedianizer();
-        medianizer.setPrice(727 * WAD);
-        medianizer.setHas(true);
+        medianizer = PipLike(OsmLike(PIP_ETH).src());
+
+        // Get current price
+        vm.prank(PAUSE_PROXY);
+        medianizer.kiss(address(this));
```
sunbreak1211, line 57:
```suggestion
        vm.prank(PAUSE_PROXY); medianizer.kiss(address(this));
```

src/tests/OracleWrapper.t.sol
```diff
@@ -17,52 +17,65 @@
...
+        medianizerPrice = uint256(medianizer.read());
+        assertGt(medianizerPrice, 0);
+        vm.prank(PAUSE_PROXY);
+        medianizer.diss(address(this));
```
sunbreak1211, line 61:
```suggestion
        vm.prank(PAUSE_PROXY); medianizer.diss(address(this));
```

src/tests/OracleWrapper.t.sol
```diff
@@ -17,52 +17,65 @@
...
+        DssInstance memory dss = MCD.loadFromChainlog(LOG);
+        assertEq(dss.chainlog.getAddress("ORACLE_WRAPPER"), address(oracleWrapper));
     }
 
-    function testReadInvalidPrice() public {
-        medianizer.setHas(false);
-        vm.expectRevert();
-        oracleWrapper.read();
+    function testRead() public {
+        assertEq(oracleWrapper.read(), bytes32(medianizerPrice/ 1800));
```
sunbreak1211, line 78:
```suggestion
        assertEq(oracleWrapper.read(), bytes32(medianizerPrice / 1800));
```

src/tests/OracleWrapper.t.sol
```diff
@@ -17,52 +17,65 @@
...
+        // Get current price
+        vm.prank(PAUSE_PROXY);
+        medianizer.kiss(address(this));
+        medianizerPrice = uint256(medianizer.read());
+        assertGt(medianizerPrice, 0);
+        vm.prank(PAUSE_PROXY);
+        medianizer.diss(address(this));
 
-        oracleWrapper = PipLike(address(new OracleWrapper(address(medianizer), address(this), 1800)));
+        oracleWrapper = PipLike(FlapperDeploy.deployOracleWrapper(address(medianizer), address(this), 1800));
```
telome, line 63: Shall we use 1200 to match the likely value that will be used in the spell?

src/deploy/FlapperInit.sol
oldchili: `want` can in theory be larger than `WAD`.
The idea was adding this for safety protection, in which case a too high `want` is not a safety concern.

src/tests/FlapperUniV2.t.sol
oldchili: Replaced with assert

src/tests/OracleWrapper.t.sol
oldchili: I think it's find to have it different. Up to Maker governance to decide.

src/tests/OracleWrapper.t.sol
```diff
@@ -17,52 +17,65 @@
...
     }
 
-    function testRead() public {
-        assertEq(oracleWrapper.read(), bytes32(727 * WAD / 1800)); // 0.40388889e+18
+    function testInitsChainlogValue() public {
+        DssInstance memory dss = MCD.loadFromChainlog(LOG);
+        assertEq(dss.chainlog.getAddress("ORACLE_WRAPPER"), address(oracleWrapper));
     }
 
-    function testReadInvalidPrice() public {
```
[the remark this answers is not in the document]
oldchili: Done.

src/deploy/FlapperInit.sol
oldchili: I meant that I can imagine having a new flapper, with a modified `initFlapper` function.
In the spell where that new flapper is changed `vow.hump` should not be touched. That's why I took it out.
Let me know if you feel strongly about that.

src/tests/FlapperMom.t.sol
oldchili: Done (though it's all random values and the initFlapperUniV2() is called only to set the chief as the mom authority).

src/deploy/FlapperInit.sol
sunbreak1211: Hmmm fair enough, I thought more the idea was to force to a value which will potentially work. But in some weird situation a value of > WAD could also do.

src/deploy/FlapperInit.sol
sunbreak1211: Hmm `bump` could also not change and we still have it in the general `init` function. It is true that `hump` might be a bit less probably that changes, but yeah just not sure, I guess the other option is to include them and spell crafter could pass the actual values.

oldchili: Ok, merged it to `initFlapperUniV2`.
I guess it's ok as the spell tests also check vow.hump value, so unlikely it will be changed to an unexpected value.

sunbreak1211 (approved):
LGTM

telome (approved):
LGTM

PR #8, opened 2023-06-27
Init direct oracle

oldchili:


sunbreak1211 (approved):
Looks good.

deploy/FlapperInit.sol
```diff
@@ -64,6 +72,12 @@ library FlapperInit {
         require(flapper.daiJoin() == address(dss.daiJoin), "Flapper daiJoin mismatch");
         require(flapper.spotter() == address(dss.spotter), "Flapper spotter mismatch");
 
+        PairLike pair = PairLike(flapper.pair());
+        (address pairDai, address pairGem) = pair.token0() == address(dss.dai) ? (pair.token0(), pair.token1())
+                                                                               : (pair.token1(), pair.token0());
+        require(pairDai == address(dss.dai), "Dai mismatch");
```
telome, line 78: Did you mean this? Otherwise the check seems redundant with the previous line.
```suggestion
        require(pairDai == flapper.dai(), "Dai mismatch");
```

telome (approved):
Minor comment, otherwise LGTM

deploy/FlapperInit.sol
oldchili: I think it's not redundant as it could be that DAI is not one of the pair tokens at all.
We don't need to check for `flapper.dai()` as that is covered in the `daiJoin` check.

telome: Ah yes my bad, you're right.

PR #10, opened 2023-07-20
Add FlapperUniV2SwapOnly

sunbreak1211:


src/FlapperUniV2.sol
```diff
@@ -35,7 +35,6 @@ interface SpotterLike {
 interface GemLike {
     function decimals() external view returns (uint8);
     function balanceOf(address) external view returns (uint256);
-    function approve(address, uint256) external;
```
oldchili, line 38: If we want to remove unused staff then we can also remove the `live()` from `VatLike`.
I'm not 100% sure these are worth changing that contract and differing from the deployed version.

src/FlapperUniV2SwapOnly.sol
```diff
@@ -0,0 +1,184 @@
...
+    function _getReserves() internal returns (uint256 reserveDai, uint256 reserveGem) {
+        (uint256 _reserveA, uint256 _reserveB,) = pair.getReserves();
+        (reserveDai, reserveGem) = daiFirst ? (_reserveA, _reserveB) : (_reserveB, _reserveA);
+
+        uint256 _daiBalance = GemLike(dai).balanceOf(address(pair));
+        uint256 _gemBalance = GemLike(gem).balanceOf(address(pair));
+        if (_daiBalance > reserveDai || _gemBalance > reserveGem) {
+            pair.sync();
+            (reserveDai, reserveGem) = (_daiBalance, _gemBalance);
+        }
```
oldchili, line 145: We should consider removing this.

src/FlapperUniV2SwapOnly.sol
```diff
@@ -0,0 +1,183 @@
...
+
+interface PipLike {
+    function read() external view returns (bytes32);
+}
+
+// https://github.com/Uniswap/v2-core/blob/ee547b17853e71ed4e0101ccfd52e70d5acded58/contracts/UniswapV2Pair.sol
+interface PairLike {
+    function getReserves() external view returns (uint112 reserve0, uint112 reserve1, uint32 blockTimestampLast);
+    function token0() external view returns (address);
+    function mint(address to) external returns (uint256 liquidity);
```
oldchili, line 47: `mint` can also be removed.

src/FlapperUniV2SwapOnly.sol
```diff
@@ -0,0 +1,183 @@
+// SPDX-FileCopyrightText: © 2023 Dai Foundation <www.daifoundation.org>
```
oldchili, line 1: When finalizing we need to add this contract to the README.

src/FlapperUniV2SwapOnly.sol
sunbreak1211: Done.

src/FlapperUniV2SwapOnly.sol
sunbreak1211: Done.

src/FlapperUniV2SwapOnly.sol
sunbreak1211: Done.

README.md
```diff
@@ -11,6 +11,15 @@ Configurable Parameters:
 * `pip` - A reference price oracle, used for bounding the exchange rate of the swap.
 * `want` - Relative multiplier of the reference price to insist on in the swap. For example, a value of 0.98 * `WAD` allows for a 2% worse price than the reference.
 
+### FlapperUniV2SwapOnly
```
oldchili, line 14: Be aware that there are notes at the end that are only relevant to the current flapper.
Maybe we should just change "also withdraws"  to "may also withdraw".

sunbreak1211: Moved things a bit around, let me know what you think.

oldchili: Yeah ok, this was the other alternative I had in mind.
Works as well.

oldchili (approved):
LGTM

oldchili (approved):
Last small change also good.

test/FlapperUniV2SwapOnly.t.sol
```diff
@@ -0,0 +1,416 @@
...
+    address     MKR;
+    address     USDC;
+    address     LINK;
+    address     PAUSE_PROXY;
+    VatLike     vat;
+    VowLike     vow;
+    EndLike     end;
+    SpotterLike spotter;
+
+    address     constant  LOG           = 0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F;
```
telome, line 109:
```suggestion
    address constant LOG                 = 0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F;
```

telome: (And the same nit in test/FlapperUniV2.t.sol)

telome (dismissed):
LGTM!

Just a small note regarding the removal of the insanity check: it could be argued that in addition to protecting the SB from being drained in a scenario of low liquidity + faulty oracle and/or incorrectly configured `want`, the check was also setting a hard cap on the swap slippage under these failure modes. So technically we are losing this protection here but I don't think it's a problem as the `hop` & `lot` combo makes sure that such failure can only have limited impact before it can be caught and mitigated.

test/FlapperUniV2SwapOnly.t.sol
sunbreak1211: Fixed in both files!

oldchili (approved):
LGTM

PR #12, opened 2023-11-29
Add Splitter

telome:


.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "lib/dss-test"]
 	path = lib/dss-test
 	url = https://github.com/makerdao/dss-test
+[submodule "lib/endgame-toolkit"]
+	path = lib/endgame-toolkit
+	url = https://github.com/dewiz-xyz/endgame-toolkit
```
sunbreak1211, line 6: I'd suggest avoiding this dependency and maybe have a mock or just the farm copied as an util or so.

src/Splitter.sol
```diff
@@ -0,0 +1,113 @@
...
+
+    // Note: the kick() cadence is determined by flapper.hop()
+    function kick(uint256 tot, uint256) external auth returns (uint256) {
+        vat.move(msg.sender, address(this), tot);
+
+        uint256 lot = tot * burn / WAD;
+        flapper.kick(lot, 0);
+
+        FarmLike farm_ = farm;
+        uint256 pay = (tot - lot) / RAY;
```
oldchili, line 103: Not sure if we should call the farming logic if `pay == 0`.
Need to test it but I think it would just extend to a new duration with nothing new to pay.

src/Splitter.sol
```diff
@@ -0,0 +1,113 @@
...
+        DaiJoinLike(daiJoin).exit(address(farm_), pay);
+        farm_.notifyRewardAmount(pay);
```
oldchili, line 105: Need to think how the farm's rewards duration can be changed (for example to align to hop), as you can't change it while a program is running. In our case it could be that the spell wants to call `setDuration` but can not execute because `notifyRewardAmount` (called from the Splitter) starves it.

There is a discussion here in the context of the other farms - https://github.com/cantinasec/review-makerdao-endgametoolkit/issues/13

I think a solution that does not require changing the farm code (but would cost a storage read) is to have a `pendingDuration` variable in the Splitter that blocks calling `notifyRewardAmount` if it's not 0.
So we check this variable, if it's 0 we call `notifyRewardAmount` regularly, and if it's not 0 we check if we can call `setDuration` (and reset it to 0), but if we can't do that we don't proceed to `notifyRewardAmount`.

We can think of other solutions as well.

src/Splitter.sol
```diff
@@ -0,0 +1,113 @@
...
+}
+
+contract Splitter {
+    mapping (address => uint256) public wards;
+
+    VatLike     public immutable vat;
+    DaiJoinLike public immutable daiJoin;
+    FarmLike    public immutable farm;
+
+    FlapLike public flapper;
```
oldchili, line 46: nit - I think we lately tend to put all the storage vars together.

deploy/FlapperInit.sol
```diff
@@ -85,16 +103,13 @@ library FlapperInit {
...
-        flapper.rely(address(dss.vow));
         flapper.rely(address(mom));
 
-        dss.vow.file("flapper", address(flapper));
         dss.vow.file("hump",    cfg.hump);
         dss.vow.file("bump",    cfg.bump);
 
         mom.setAuthority(dss.chainlog.getAddress("MCD_ADM"));
 
-        dss.chainlog.setAddress("MCD_FLAP",    address(flapper));
```
oldchili, line 97: I think the low level flapper (burner) should have a representation in the chainlog, as it is authed by the PP (and its `hop` parameter basically sets the cadence of the whole thing).

src/Splitter.sol
telome: Good point. The issue with the `pendingDuration` solution is that the farm has only a single owner (which would be the PauseProxy) so the Splitter would not be able to directly call `setRewardsDuration` (without adding an intermediary owner contract). 

We could also just set `burn = WAD` (and add the condition you mentioned above to skip calling the farm when we burn 100%), but that means the change to the duration would require two spells. Maybe that's still better than implementing a dedicated farm owner contract?

deploy/FlapperInit.sol
telome: What do you guys think of keeping `MCD_FLAP` for whatever is set as `flapper` in the `vow` and use more specific names for all potential flappers / flapper modules (e.g. `MCD_FLAP_SPLIT` for the splitter, `MCD_FLAP_LP` for the current flapper and `MCD_FLAP_BURN` for the future burner)?

src/Splitter.sol
oldchili: Ah interesting about the single owner.
I think two spells should really be a last resort (as it involves a lot of coordination and explaining) and a dedicated owner contract seems a bit ugly indeed.

So think we can change the farm contract for something like this, but better have a single version with endgame-toolkit, so ideally it's a change we can push there. Note that Dewiz already added a reference param there to one of the functions, so this version was already changed by Maker.

One option is to just support multiple owners (though it's a big ugly with the inheritance).

Another option is to change `setRewardsDuration` to be called either by `owner` or by `rewardsDistribution`. For that we need to remove the `onlyOwner` modifier from it and add something like:
`require(msg.sender == owner || msg.sender == rewardsDistribution);`
This imo is a bit nicer sine we extend the special role that already exists, and don't give the splitter a full `owner` status (though some would say that would be more Maker-like).

Finally, we can consider doing the change suggested by Cantina, as it seems that also the regular farms might benefit from it. But that needs to be done carefully, with a thorough review and audit ofc.

deploy/FlapperInit.sol
oldchili: will defer to @sunbreak1211 for this one :)

src/Splitter.sol
telome: Changed

src/Splitter.sol
telome: Fixed

.gitmodules
telome: Changed

deploy/FlapperInit.sol
```diff
@@ -56,14 +69,19 @@ struct FlapperUniV2Config {
...
+struct SplitterConfig {
+    address splitter;
+    uint256 burn;
+}
+
 library FlapperInit {
     uint256 constant WAD = 10 ** 18;
 
-    function initFlapperUniV2(
+    function _initFlapperUniV2(
```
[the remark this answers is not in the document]
telome: Done (pending a decision regarding the chainlog keys).

deploy/FlapperInit.sol
```diff
@@ -106,4 +121,47 @@ library FlapperInit {
...
+
+        // Sanity checks
+        SplitterLike splitter = SplitterLike(splitterCfg.splitter);
+        require(splitter.vat()     == address(dss.vat),     "Splitter vat mismatch");
+        require(splitter.daiJoin() == address(dss.daiJoin), "Splitter daiJoin mismatch");
+
+        require(splitterCfg.burn <= WAD, "Splitter burn too high");
+
+        splitter.file("burn", splitterCfg.burn);
+        FarmLike(splitter.farm()).setRewardsDistribution(address(splitter));
```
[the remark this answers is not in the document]
telome: Done

.gitmodules
oldchili: Are we sure about this?
The nice think about having the dependency is that if we make changes in the endgame-toolkit code (which we are now discussing) we will have them here just by updating the dependency.
Also, there are some small changes in the mock that it might be nicer not to test with.

deploy/FlapperInit.sol
oldchili: I think the issue is that we are trying to support both a mode without a splitter and a mode with one. So not sure how removing `MCD-FLAP` fits into that. Maybe if we do that as part of `initSplitter`.

Regardless, note that in the current code we don't really need the setting of `MCD_FLAP` explicitly in `initFlapperUniV2()` and overriding in `initSplitter()`, as even if we are just using a flapper without a splitter we can pass `MCD_FLAP` as `cfg.chainlogKey`.

src/Splitter.sol
telome: Oh yes, done.

src/Splitter.sol
telome: Created this forked branch with the farm change to `setRewardDuration()` here https://github.com/makerdao/dss-flappers/pull/13. If we agree that this is the way to go, we can port those changes to [StakingRewards.sol](https://github.com/dewiz-xyz/endgame-toolkit/blob/master/src/synthetix/StakingRewards.sol)

src/Splitter.sol
```diff
@@ -0,0 +1,114 @@
...
+    // Note: the kick() cadence is determined by flapper.hop()
+    function kick(uint256 tot, uint256) external auth returns (uint256) {
+        vat.move(msg.sender, address(this), tot);
+
+        uint256 lot = tot * burn / WAD;
+        flapper.kick(lot, 0);
+
+        if (tot - lot >= RAY) {
+            FarmLike farm_ = farm;
+            uint256 pay = (tot - lot) / RAY;
```
sunbreak1211, line 103: nit:
```suggestion
        uint256 pay = (tot - lot) / RAY;
        if (pay >= 0) {
            FarmLike farm_ = farm;
```

src/Splitter.sol
```diff
@@ -0,0 +1,114 @@
...
+
+    // Note: the kick() cadence is determined by flapper.hop()
+    function kick(uint256 tot, uint256) external auth returns (uint256) {
+        vat.move(msg.sender, address(this), tot);
+
+        uint256 lot = tot * burn / WAD;
+        flapper.kick(lot, 0);
+
+        if (tot - lot >= RAY) {
+            FarmLike farm_ = farm;
```
sunbreak1211, line 102: `farm` is immutable, so we do not need the cache. However, any reason to not make it manageable as the flapper? I do not see a real problem being able to change it.

src/Splitter.sol
telome: I think dust can only accumulate here if `tot` is not a multiple of `RAD`, which seems like an unlikely setting. So I'd say it's fine to allow dust to theoretically accumulate for weird `bump` values that would not be multiples of `RAD`.

src/Splitter.sol
telome: Good catch.

The probability of having to do a farm migration just seems quite low so it seems fine having to redeploy a new Splitter for that. That's in contrast with the requirement of being able to change the flapper between FlapperUniV2 and FlapperUniV2SwapOnly according to the whims of governance.

src/Splitter.sol
sunbreak1211: A multiple of `RAY`, right?
It would also affect the value of `burn` to this, but yeah I guess it is not a big deal. Anyone could also send `vat.dai` and we would be in a similar situation. Pause proxy can actually remove this balance if needed as well.

telome: No I meant `RAD` so that this is independent of the `burn` value. If `tot = t * RAD` then `tot - tot * burn / WAD = t * RAD - t * RAD * burn / WAD = t * RAD - t * RAY * burn`, which is a multiple of `RAY` for any values of the `t` and `burn` integers.

foundry.toml
```diff
@@ -7,4 +7,8 @@ optimizer = true
 optimizer_runs = 200
 verbosity = 3
 
+remappings = [
+    "openzeppelin-contracts/=lib/endgame-toolkit/lib/openzeppelin-contracts/contracts/"
+]
+
```
sunbreak1211, line 13:
```suggestion
```

test/FlapperMom.t.sol
```diff
@@ -82,7 +84,9 @@ contract FlapperMomTest is Test {
             want: 1e18,
             pip:  address(0),
             hump: 1,
-            bump: 0
+            bump: 0,
+            caller: VOW,
+            chainlogKey: "MCD_FLAP_LP"
```
sunbreak1211, line 89: Shouldn't just be better to do `MCD_FLAP` here?

foundry.toml
telome: Done

deploy/FlapperInit.sol
```diff
@@ -106,4 +133,28 @@ library FlapperInit {
...
+    function initSplitter(        
+        DssInstance memory dss,
+        address splitter,
+        SplitterConfig memory cfg
+    ) internal {
+        // Sanity checks
+        SplitterLike splitter_ = SplitterLike(splitter);
+        require(splitter_.vat()     == address(dss.vat),     "Splitter vat mismatch");
+        require(splitter_.daiJoin() == address(dss.daiJoin), "Splitter daiJoin mismatch");
+
```
oldchili, line 146: This is something that we need to decide our policy about but I'm wondering if we should add a sanity check that `splitter_.farm()` matches a `cfg.farm` that we get as a paramter.

These immutables deployer-integrity checks are pretty annoying but we got a similar issue once (about not validating `univ3Factory`). Here allegedly a malicious farm can still the 30% of streamed dai.

And while we're at it if we do that we should probably add similar checks for the flapper `pair` and `receiver` in `initFlapperUniV2`.

test/Splitter.t.sol
```diff
@@ -0,0 +1,408 @@
...
+
+        vm.expectRevert("Previous rewards period must be complete before changing the duration for the new period");
+        vm.prank(PAUSE_PROXY); farm.setRewardsDuration(666 minutes); 
+
+        vm.prank(PAUSE_PROXY); splitter.file("burn", WAD);
+
+        vm.warp(block.timestamp + farm.rewardsDuration() + 1);
+        doKick();
+
+        vm.prank(PAUSE_PROXY); farm.setRewardsDuration(666 minutes); // should not revert
```
oldchili, line 373: I think we should add:

-A test for increasing the rewards duration and hop while there is a farming program. If I'm not mistaking we will have one intermediate period with a lower reward rate (between the setDuration/file(hop) and the end of the new hop), then should start the new cadence with the new reward rate.

-A test for increasing the rewards duration and hop while there is a farming program. Similarly we should have one intermediate period with a higher reward rate.

test/mocks/StakingRewardsMock.sol
```diff
@@ -0,0 +1,212 @@
+/**
+ * NOTICE: This contract has been adapted from the original Synthetix source code:
+ *  - Upgrade Solidity version from 0.5.x to 0.8.x.
+ *  - Add referral code functionality for `stake()`.
+ *
+ * Original: https://github.com/Synthetixio/synthetix/blob/5e9096ac4aea6c4249828f1e8b95e3fb9be231f8/contracts/StakingRewards.sol
+ *     Diff: https://www.diffchecker.com/rgBMl45S/
+ */
+
```
sunbreak1211, line 9:
```suggestion
```
I'd remove this as not really important for a mock, also because the diff is old and nor appropriate for it, as it is not including the specific mock changes.

deploy/FlapperInit.sol
telome: Ok I guess it doesn't hurt to add those. Done.

deploy/FlapperInit.sol
```diff
@@ -106,4 +138,29 @@ library FlapperInit {
...
+        // Sanity checks
+        SplitterLike splitter_ = SplitterLike(splitter);
+        require(splitter_.vat()     == address(dss.vat),     "Splitter vat mismatch");
+        require(splitter_.daiJoin() == address(dss.daiJoin), "Splitter daiJoin mismatch");
+        require(splitter_.farm()    == cfg.farm,             "Splitter farm mismatch");
+
+        require(cfg.burn <= WAD, "Splitter burn too high");
+
+        splitter_.file("burn", cfg.burn);
+        FarmLike farm = FarmLike(splitter_.farm());
```
sunbreak1211, line 156:
```suggestion
        FarmLike farm = FarmLike(cfg.farm);
```
To avoid the extra SLOAD.

test/FlapperMom.t.sol
```diff
@@ -78,11 +80,14 @@ contract FlapperMomTest is Test {
...
-            hump: 1,
-            bump: 0
+            hop:          5 minutes,
+            want:         1e18,
+            pip:          address(0),
+            hump:         1,
+            bump:         0,
+            pair:         UNIV2_DAI_MKR_PAIR,
+            caller:       VOW,
+            chainlogKey: "MCD_FLAP"
```
sunbreak1211, line 90:
```suggestion
            hop:                5 minutes,
            want:              1e18,
            pip:                 address(0),
            hump:             1,
            bump:             0,
            pair:                UNIV2_DAI_MKR_PAIR,
            caller:             VOW,
            chainlogKey: "MCD_FLAP"
```

sunbreak1211, line 90: All the values got shifted one extra space compared to the last one. I tried to send a suggestion with the fix but was doing some weird stuff. There might be some spaces vs tab issue in the file (I think the default is we make 4 spaces and make sure there is not any tab lost around there).

telome: Good eyes :) Should be fixed now. I double checked that there is no leftover tab anywhere as well.

test/Splitter.t.sol
telome: Yes, you're right, done.

src/Splitter.sol
oldchili: This is being worked on [here](https://github.com/makerdao/endgame-toolkit/pull/14) so I'm going to resolve.

test/Splitter.t.sol
oldchili: nicely done

src/Splitter.sol
```diff
@@ -0,0 +1,113 @@
...
+        vat.move(msg.sender, address(this), tot);
+
+        uint256 lot = tot * burn / WAD;
+        flapper.kick(lot, 0);
+
+        FarmLike farm_ = farm;
+        uint256 pay = (tot - lot) / RAY;
+        DaiJoinLike(daiJoin).exit(address(farm_), pay);
+        farm_.notifyRewardAmount(pay);
+
```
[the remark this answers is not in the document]
oldchili: @sunbreak1211 do we have a policy that such function should add an event or can we skip it here?

test/Splitter.t.sol
```diff
@@ -0,0 +1,433 @@
...
+        // Add initial liquidity if needed
+        (uint256 reserveDai, ) = UniswapV2Library.getReserves(UNIV2_FACTORY, DAI, MKR);
+        uint256 minimalDaiReserve = 280_000 * WAD;
+        if (reserveDai < minimalDaiReserve) {
+            changeUniV2Price(medianizer.read(), MKR, UNIV2_DAI_MKR_PAIR);
+            (reserveDai, ) = UniswapV2Library.getReserves(UNIV2_FACTORY, DAI, MKR);
+            if(reserveDai < minimalDaiReserve) {
+                topUpLiquidity(minimalDaiReserve - reserveDai, MKR, UNIV2_DAI_MKR_PAIR);
+            }
+        }
```
sunbreak1211, line 178: Shouldn't be doing something similar to the other tests that we ensure the prices get aligned. It's built a bit different and it is true that if resetveDai < minimalDaiReverve prices will get aligned, but won't be the case with the actual pool deposits, meaning that if the oracle gets outdated the tests will start failing. I might be missing something that comes later though.

.gitmodules
sunbreak1211: I don't really like having an external dependency here. Makes the repos heavier and in future we might start having incompatibility issues due to solidity versions. If we do not want to test with a mock we might prefer to copy the real thing.

test/Splitter.t.sol
```diff
@@ -0,0 +1,433 @@
...
+
+        assertEq(vat.can(address(splitter), address(0xf1)), 0);
+        assertEq(vat.can(address(splitter), address(0xf2)), 1);
+    }
+
+    function testKick() public {
+        doKick();
+    }
+
+    function testKickBurnOnly() public {
```
sunbreak1211, line 321: Shouldn't this test just be the config of burn value and then doKick?

deploy/FlapperDeploy.sol
```diff
@@ -24,6 +24,7 @@ import { FlapperUniV2 } from "src/FlapperUniV2.sol";
 import { FlapperUniV2SwapOnly } from "src/FlapperUniV2SwapOnly.sol";
 import { FlapperMom } from "src/FlapperMom.sol";
 import { OracleWrapper } from "src/OracleWrapper.sol";
+import { Splitter } from "src/Splitter.sol";
```
oldchili, line 27: Probably need to mention the Splitter in the README as well.

telome: Added

test/Splitter.t.sol
```diff
@@ -0,0 +1,433 @@
...
+        vow.flap();
+    }
+
+    function checkChangeRewardDuration(uint256 newDuration) private {
+        uint256 topup = 5707 * (WAD - 70 * WAD / 100);
+        doKick();
+        assertEq(farm.rewardsDuration(), flapper.hop());
+        assertEq(farm.rewardsDuration(), 30 minutes);
+        assertEq(farm.rewardRate(), topup / 30 minutes);
+        uint256 prevRewardRate = farm.rewardRate();
```
sunbreak1211, line 366:
```suggestion
        uint256 prevRewardRate = farm.rewardRate();
        assertEq(prevRewardRate, topup / 30 minutes);
```

.gitmodules
oldchili: Yeah still not sure we agree here, but not a hill I would die on.
Resolving.

test/Splitter.t.sol
telome: Slight preference to keeping it as-is as semantically the reward rate is not yet the "previous" reward rate when it is being checked here.

test/Splitter.t.sol
telome: You're right, fixed.

test/Splitter.t.sol
telome: Yes, I think it's indeed better not to rely on the oracle being up-to-date. Changed.

src/Splitter.sol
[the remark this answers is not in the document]
sunbreak1211: I don't think we have a policy for this and it is true that the info is already saved samewhere else but I slightly tend to think we could add it, but really not strong feelings about this.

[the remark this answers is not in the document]
telome: Added an event so it's a bit more consistent with the other flappers.

README.md
```diff
@@ -34,6 +34,14 @@ This contract allows bypassing the governance delay when disabling the Flapper i
 
 Allows for scaling down an oracle price by a certain value. This can be useful when the `gem` is a redenominated version of an existing token, which already has a reliable oracle.
 
+### Splitter
+
+Exposes a `kick` operation to be triggered periodically. Its logic withdraws `DAI` from the `vow` and splits it in two parts. The first part (`burn`) is sent to the underlying `flapper` contract to be burned. The second part (`WAD - burn`) is distributed as reward to a `farm` contract. The `kick` cadence is determined by `flapper.hop()`.
```
oldchili, line 39:
```suggestion
Exposes a `kick` operation to be triggered periodically. Its logic withdraws `DAI` from the `vow` and splits it in two parts. The first part (`burn`) is sent to the underlying `flapper` contract to be processed by the burn engine. The second part (`WAD - burn`) is distributed as reward to a `farm` contract. The `kick` cadence is determined by `flapper.hop()`.
```

README.md
```diff
@@ -34,6 +34,14 @@ This contract allows bypassing the governance delay when disabling the Flapper i
...
+
+Configurable Parameters:
+* `flapper` - The underlying burner strategy (e.g. the address of `FlapperUniV2SwapOnly`).
+* `burn` - The percentage of the `vow.bump` to be moved to the underlying `flapper`. For example, a value of 0.70 \* `WAD` corresponds to a 70% burn.
```
oldchili, line 43:
```suggestion
* `burn` - The percentage of the `vow.bump` to be moved to the underlying `flapper`. For example, a value of 0.70 \* `WAD` corresponds to funneling 70% of the `DAI` to the burn engine.
```

src/Splitter.sol
```diff
@@ -0,0 +1,115 @@
...
+}
+
+interface FarmLike {
+    function notifyRewardAmount(uint256 reward) external;
+}
+
+contract Splitter {
+    mapping (address => uint256) public wards;
+    FlapLike    public           flapper;
+    uint256     public           burn; // in WAD; 1 WAD = 100% burn
```
oldchili, line 42:
```suggestion
    uint256     public           burn; // in WAD; 1 WAD = funneling 100% to the burn engine
```

src/Splitter.sol
```diff
@@ -0,0 +1,115 @@
...
+        else revert("Splitter/file-unrecognized-param");
+        emit File(what, data);
+    }
+
+    // Note: the kick() cadence is determined by flapper.hop()
+    function kick(uint256 tot, uint256) external auth returns (uint256) {
+        vat.move(msg.sender, address(this), tot);
+
+        uint256 lot = tot * burn / WAD;
+        flapper.kick(lot, 0);
```
oldchili, line 100: Something we should be aware of is that there could be a case where we might want `burn == 0`.
In that case the expected behavior would be to send everything to the farm, but still enforce the rate limit.
Currently `FlapperUniv2` reverts in this case (getting "revert: UniswapV2: INSUFFICIENT_OUTPUT_AMOUNT").

Probably the minimum we should do is comment that a very low `burn` value will cause the flapper to fail and is not supported. But in general this is a point in favor of moving the rate enforcement to the Splitter (which is probably not something we want to do yet).

telome: Added a note to the README and a test checking the splitter around the minimum required burn value for `FlapperUniV2SwapOnly`.

oldchili: Ok, I'm still not 100% at ease with not supporting `burn == 0`, but it might be good enough for this version.
@sunbreak1211 any opinion here?

deploy/FlapperInit.sol
```diff
@@ -106,4 +138,29 @@ library FlapperInit {
...
+        address splitter,
+        SplitterConfig memory cfg
+    ) internal {
+        // Sanity checks
+        SplitterLike splitter_ = SplitterLike(splitter);
+        require(splitter_.vat()     == address(dss.vat),     "Splitter vat mismatch");
+        require(splitter_.daiJoin() == address(dss.daiJoin), "Splitter daiJoin mismatch");
+        require(splitter_.farm()    == cfg.farm,             "Splitter farm mismatch");
+
+        require(cfg.burn <= WAD, "Splitter burn too high");
```
oldchili, line 153:
```suggestion
        require(cfg.burn <= WAD, "Splitter burn too high");
        require(cfg.burn > 0, "Splitter burn is zero"); // Note that also other very low `burn` values may cause `flapper.kick` to revert.
```

telome: Added

oldchili (dismissed):
LGTM for merging this implementation.
We might want to discuss moving the rate-limiting to the splitter and supporting `burn==0` at some point.

oldchili:
Actually we probably want to merge this to `splitter` first - 
https://github.com/makerdao/dss-flappers/pull/14

src/Splitter.sol
sunbreak1211: This is a good question. I think we need to implement a rate limiting at the splitter level. And ofc the tighter one will be the one marking the pace.
In the case of burn == 0, we just do not call the Flapper.
Is there a real reason to not do this already? the extra storage read?

sunbreak1211:
> Actually we probably want to merge this to `splitter` first - #14

I'm gonna look at this soon as I haven't even glanced that PR yet.

src/Splitter.sol
sunbreak1211: Yeah to be honest I think the most clean way to do this would be having always the splitter and having the rate limiter there. Then if it should only go to the flapper configured for that. For sure is also a gas waste that configuration.

oldchili: Yes, assuming we change the current implementation I also prefer to have the rate limit only in the Splitter than having it in both places.

telome: Ok, for reference [here](https://github.com/makerdao/dss-flappers/pull/12/commits/0e64c8816e0530e9bffa657a233f8383b3a8553a) is an implementation where we have rate-limit in both the Splitter and Flapper but set the flapper.hop() to zero to indicate that the flapper rate limit is effectively not used. 

Next, I'll add a commit with the flapper rate-limiting removed (less wasteful but less generic) and I guess we can decide which one we like best.

test/mocks/StakingRewardsMock.sol
```diff
@@ -0,0 +1,203 @@
+// SPDX-FileCopyrightText: © 2019-2021 Synthetix
```
sunbreak1211, line 1: Let's update this one with the latest version.

deploy/FlapperInit.sol
```diff
@@ -48,12 +49,41 @@ interface PairLike {
...
     uint256 want;
     address pip;
     uint256 hump;
     uint256 bump;
+    address pair;
+    address caller;
+    bytes32 chainlogKey;
+}
+
+struct SplitterConfig {
```
sunbreak1211, line 81: We could consider to just have the `hop` value and set directly that to the rewardsDistribution as well.

sunbreak1211:
I think we should branch off `master` in a new `dev` branch and base that PR in `dev`, so we can merge and audit there and only merge to `master` we are ready to go to production with the new set up.

src/Splitter.sol
telome: Ended up opening [another PR](https://github.com/makerdao/dss-flappers/pull/16) for the version without flapper rate-limit as the diff was quite large and I needed to build the changes on top of #14 for the tests to pass.

test/mocks/StakingRewardsMock.sol
telome: This mock is based on the latest version of https://github.com/makerdao/endgame-toolkit/blob/master/src/synthetix/StakingRewards.sol . I've just added back the notice that had been removed.

sunbreak1211: Hmm ok, this doesn't include the constructor `require` change and some other minor comment changes but yeah the PR wasn't merged yet.

telome: Ah yes right, I'll include these changes as well

src/Splitter.sol
oldchili: Resolving this comment as we now support `burn == 0`.

src/Splitter.sol
```diff
@@ -0,0 +1,124 @@
...
+
+    function kick(uint256 tot, uint256) external auth returns (uint256) {
+        require(block.timestamp >= zzz + hop, "Splitter/kicked-too-soon");
+        zzz = block.timestamp;
+
+        vat.move(msg.sender, address(this), tot);
+
+        uint256 lot = tot * burn / WAD;
+        if (lot > 0) {
+            flapper.exec(lot);
```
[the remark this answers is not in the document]
telome: Yes, done.

src/Splitter.sol
```diff
@@ -0,0 +1,121 @@
...
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+pragma solidity ^0.8.16;
+
+interface VatLike {
+    function move(address, address, uint256) external;
+    function hope(address) external;
+    function nope(address) external;
```
sunbreak1211, line 22:
```suggestion
```

src/FlapperUniV2SwapOnly.sol
```diff
@@ -33,6 +33,7 @@ interface SpotterLike {
 
 interface GemLike {
     function decimals() external view returns (uint8);
+    function transfer(address, uint256) external;
```
sunbreak1211, line 36: Can't mark in the line, but we should remove `move` from `VatLike`.

src/FlapperUniV2.sol
```diff
@@ -16,6 +16,8 @@
 
 pragma solidity ^0.8.16;
 
+import { Babylonian } from "src/Babylonian.sol";
+
 interface VatLike {
     function hope(address) external;
     function move(address, address, uint256) external;
```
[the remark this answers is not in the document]
sunbreak1211: Also remove `DaiJoinLike.exit`

src/FlapperUniV2SwapOnly.sol
sunbreak1211: And `exit` from `DaiJoinLike`

src/FlapperUniV2.sol
```diff
@@ -148,12 +145,12 @@ contract FlapperUniV2 {
...
-    // (2)   (total - lot) / bought  = (reserveDai + lot) / (reserveGem - bought)
+    // (1)   reserveDai * reserveGem = (reserveDai + sell * 997 / 1000) * (reserveGem - bought)
+    // (2)   (lot - sell) / bought  = (reserveDai + sell) / (reserveGem - bought)
     //
-    // The solution for the these equations for variable `total` and `bought` is used below.
-    function _getTotalDai(uint256 wlot, uint256 reserveDai) internal pure returns (uint256 total) {
-        total = wlot * (997 * wlot + 1997 * reserveDai) / (1000 * reserveDai);
+    // The solution for the these equations for variables `sell` and `bought` is used below.
+    function _getDaiToSell(uint256 lot, uint256 reserveDai) internal pure returns (uint256 sell) {
+        sell = (Babylonian.sqrt(reserveDai * (lot * 3988000 + reserveDai * 3988009)) - reserveDai * 1997) / 1994;
```
sunbreak1211, line 153: nit:
```suggestion
        sell = (Babylonian.sqrt(reserveDai * (lot * 3_988_000 + reserveDai * 3_988_009)) - reserveDai * 1997) / 1994;
```

deploy/FlapperInit.sol
```diff
@@ -52,30 +53,58 @@ interface DaiJoinLike {
...
     uint256 hump;
     uint256 bump;
+    uint256 hop;
+    uint256 burn;
+    address farm;
     address daiJoin;
+    bytes32 splitterChainlogKey;
```
sunbreak1211, line 87: Shouldn't we also have a prev splitter key?

telome: I think if the standard is to write an init function to support the first onboarding spell (as discussed here https://github.com/makerdao/dss-flappers/pull/16#discussion_r1452292706)  then it might be confusing to also include code for cleanup that will only hypothetically be needed at some point in the future.

deploy/FlapperInit.sol
```diff
@@ -84,24 +113,16 @@ library FlapperInit {
...
-        dss.vow.file("flapper", address(flapper));
-        dss.vow.file("hump",    cfg.hump);
-        dss.vow.file("bump",    cfg.bump);
+        SplitterLike(cfg.splitter).file("flapper", flapper_);
 
-        mom.setAuthority(dss.chainlog.getAddress("MCD_ADM"));
-
-        dss.chainlog.setAddress("MCD_FLAP",    address(flapper));
-        dss.chainlog.setAddress("FLAPPER_MOM", address(mom));
+        if (cfg.prevChainlogKey != bytes32(0)) dss.chainlog.removeAddress(cfg.prevChainlogKey);
```
sunbreak1211, line 124:
```suggestion
        if (cfg.prevChainlogKey != bytes32(0) && cfg.prevChainlogKey != cfg.chainlogKey) dss.chainlog.removeAddress(cfg.prevChainlogKey);
```

sunbreak1211: Not really sure is needed though.

deploy/FlapperInit.sol
sunbreak1211: Yeah could be, it feels a bit annoying the other contracts have the prev keys and not here but I do not have a real strong position on this anyway.

deploy/FlapperInit.sol
telome: I'd expect `cfg.prevChainlogKey` to be set to `bytes32(0)` in case an existing chainlog key is reused. Worse case if it's set to non-zero by mistake, it will just remove and re-add the address (inefficient but still correct final state).

deploy/FlapperInit.sol
```diff
@@ -112,4 +133,41 @@ library FlapperInit {
...
+        SplitterInstance memory splitterInstance,
+        SplitterConfig   memory cfg
+    ) internal {
+        SplitterLike    splitter = SplitterLike(splitterInstance.splitter);
+        SplitterMomLike mom      = SplitterMomLike(splitterInstance.mom);
+
+        // Sanity checks
+        require(splitter.vat()     == address(dss.vat), "Splitter vat mismatch");
+        require(splitter.daiJoin() == cfg.daiJoin,      "Splitter daiJoin mismatch");
+        require(splitter.farm()    == cfg.farm,         "Splitter farm mismatch");
```
sunbreak1211, line 148: We could also check `mom.splitter() == splitter`.

deploy/FlapperInit.sol
```diff
@@ -84,24 +113,16 @@ library FlapperInit {
...
         flapper.file("want", cfg.want);
         flapper.file("pip",  cfg.pip);
-        flapper.rely(address(dss.vow));
-        flapper.rely(address(mom));
+        flapper.rely(cfg.splitter);
 
-        dss.vow.file("flapper", address(flapper));
-        dss.vow.file("hump",    cfg.hump);
-        dss.vow.file("bump",    cfg.bump);
+        SplitterLike(cfg.splitter).file("flapper", flapper_);
```
sunbreak1211, line 122: I feel this line should happen in `initSplitter` instead of here.

sunbreak1211: Well actually depends on the case. For the very initial set up I feel should go in `initSplitter`. It is true that if we need to upgrade the flapper goes in this place. Probably fine as it is.

deploy/FlapperInit.sol
telome: Yes, good point, added.

README.md
```diff
@@ -1,28 +1,33 @@
 # Dss Flappers
 
-Implementations of MakerDao surplus auctions, triggered on `vow.flap`. The current featured Flapper is `FlapperUniV2`.
+Implementations of MakerDao surplus auctions, triggered on `vow.flap`.
```
oldchili, line 3: Need to update the documentation about `FlapperMom` (replace with `SplitterMom` etc..).

README.md
```diff
@@ -1,28 +1,33 @@
...
 
-### FlapperUniV2
+### Splitter
```
oldchili, line 5: "the `vow` is expected to hold at least the swapped amount" =>
"the `vow` is expected to hold at least the consumed amount"

deploy/FlapperInit.sol
telome: I guess there are merits to both ways but something that makes me lean towards keeping the filing in  `initFlapperUniV2` is that the Splitter can in principle be used without a flapper (when `burn == 0`), whereas the flapper always needs the Splitter.

src/FlapperUniV2SwapOnly.sol
telome: Done

src/FlapperUniV2.sol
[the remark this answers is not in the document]
telome: Done

src/FlapperUniV2.sol
telome: Done

README.md
```diff
@@ -1,39 +1,44 @@
 # Dss Flappers
 
-Implementations of MakerDao surplus auctions, triggered on `vow.flap`. The current featured Flapper is `FlapperUniV2`.
+Implementations of MakerDao surplus auctions, triggered on `vow.flap`.
```
sunbreak1211, line 3:
```suggestion
Implementations of MakerDAO surplus auctions, triggered on `vow.flap`.
```

test/FlapperUniV2.t.sol
```diff
@@ -299,150 +307,127 @@ contract FlapperUniV2Test is DssTest {
...
-        emit Cage(0);
+        emit Cage();
         vm.prank(PAUSE_PROXY); end.cage();
         assertEq(flapper.live(), 0);
     }
+
+    // A shortened version of the sell and deposit flapper that sells `lot`.
+    // Based on: https://github.com/makerdao/dss-flappers/blob/da7b6b70e7cfe3631f8af695bbe0c79db90e2a20/src/FlapperUniV2.sol
+    function sellLotAndDeposit(PairLike pair, address gem, bool daiFirst, address receiver, uint256 lot) internal {
+
```
sunbreak1211, line 391:
```suggestion
```

test/Splitter.t.sol
```diff
@@ -0,0 +1,455 @@
...
+        assert(splitter.wards(address(this)) == 0);
+
+        checkModifier(address(splitter), string(abi.encodePacked("Splitter", "/not-authorized")), [
+            Splitter.kick.selector,
+            Splitter.cage.selector
+        ]);
+    }
+
+    function testFileUint() public {
+        checkFileUint(address(splitter), "Splitter", ["burn"]);
```
sunbreak1211, line 315:
```suggestion
        checkFileUint(address(splitter), "Splitter", ["burn", "hop"]);
```

deploy/FlapperInit.sol
sunbreak1211: Resolving this as I mention I do not have a real strong position here.

oldchili (approved):
LGTM

PR #13, opened 2023-12-06
Update farm's setRewardsDuration logic

telome:


test/mocks/StakingRewardsMock.sol
```diff
@@ -165,11 +165,15 @@ contract StakingRewardsMock {
...
-        require(
-            block.timestamp > periodFinish,
-            "Previous rewards period must be complete before changing the duration for the new period"
-        );
+    function setRewardsDuration(uint256 _rewardsDuration) external onlyOwner updateReward(address(0)) {
+        uint256 periodFinish_ = periodFinish;
+        if (block.timestamp < periodFinish_) {
+            uint256 leftover = (periodFinish_ - block.timestamp) * rewardRate;
+            rewardRate = leftover / _rewardsDuration;
+            lastUpdateTime = block.timestamp;
```
oldchili, line 173: Are you sure we need this? 
Inside `updateReward` we have `lastUpdateTime = lastTimeRewardApplicable();` and in `lastTimeRewardApplicable` we have:
`return block.timestamp < periodFinish ? block.timestamp : periodFinish;`

telome: Indeed it's not necessary, I only kept it to keep the logic more similar with `notifyRewardAmount()` so that it's more immediately obvious that the code in that `if` is basically just doing `notifyRewardAmount(0);`. But thinking about it now I think it's probably cleaner to just remove the line.

telome:
Closing as the change to the mock was included in ee15c52

PR #15, opened 2024-01-08
Support also NST on init validation

oldchili:


telome (approved):
LGTM. I'll add that change to `initSplitter` in https://github.com/makerdao/dss-flappers/pull/12 as well after this is merged.

sunbreak1211:
> LGTM. I'll add that change to `initSplitter` in #12 as well after this is merged.

Still need to look at it. But what I'd do is after merging this to `master`, branch off `master` again to a new `dev`. Then we can base the splitter PR to that branch, merge this last change into the splitter and do the splitter change. That should be probably a good way to keep it ordered.

sunbreak1211 (approved):
LGTM

PR #17, opened 2024-02-02
Splitter audit fix

telome:


src/Splitter.sol
```diff
@@ -38,6 +38,7 @@ interface FarmLike {
 contract Splitter {
     mapping (address => uint256) public wards;
     FlapLike    public           flapper;
+    uint256     public           live; // Active Flag
```
sunbreak1211, line 41: Not really with a strong reason but I'd put the `live` flag right after `wards`.

src/Splitter.sol
```diff
@@ -115,6 +121,7 @@ contract Splitter {
     }
 
     function cage(uint256) external auth {
```
sunbreak1211, line 123: Wondering if we should still `cage` the `flap` in case in the future is necessary to do some settlement.

deploy/FlapperInit.sol
```diff
@@ -92,6 +87,7 @@ struct SplitterConfig {
 
 library FlapperInit {
     uint256 constant WAD = 10 ** 18;
+    uint256 constant RAY = 10 ** 45;
```
sunbreak1211, line 90:
```suggestion
    uint256 constant RAD = 10 ** 45;
```

deploy/FlapperInit.sol
```diff
@@ -150,6 +144,7 @@ library FlapperInit {
         require(mom.splitter()     == splitterInstance.splitter, "Mom splitter mismatch");
 
         require(cfg.hump > 0,         "hump too low");
+        require(cfg.bump % RAY == 0,  "bump not multiple of RAY");
```
sunbreak1211, line 147:
```suggestion
        require(cfg.bump % RAD == 0,  "bump not multiple of RAD");
```

deploy/FlapperInit.sol
sunbreak1211: This comes from the audit, they should change it there. Also they mention `hump` instead of `bump` in the description.

src/Splitter.sol
telome: Done

deploy/FlapperInit.sol
telome: Ah I actually meant RAY here. I think RAY is correct as the no-dust condition is `(tot/RAY - lot + lot) * RAY == tot` which is satisfied if `tot` is a multiple of `RAY`.

src/FlapperUniV2.sol
```diff
@@ -73,35 +61,29 @@ contract FlapperUniV2 {
...
-        dai = daiJoin.dai();
+        dai = _dai;
         gem = _gem;
         require(GemLike(gem).decimals() == 18, "FlapperUniV2/gem-decimals-not-18");
 
         pair     = PairLike(_pair);
         daiFirst = pair.token0() == dai;
         receiver = _receiver;
 
-        vat.hope(address(daiJoin));
```
oldchili, line 97: redundant empty line here

src/Splitter.sol
oldchili: Can you give an example of what you mean by settlement?
Is it for a future case where the flapper also hold funds?

deploy/FlapperInit.sol
sunbreak1211: Ah right, resolving.

src/Splitter.sol
sunbreak1211: Yeah as discussed, this is thinking on a future Flapper that might hold funds. Anyway it seems we don't think is necessary to do it, so feel free to solve this one.

src/Splitter.sol
```diff
@@ -37,10 +37,11 @@ interface FarmLike {
 
 contract Splitter {
     mapping (address => uint256) public wards;
-    FlapLike    public           flapper;
-    uint256     public           burn; // [WAD]       Burn percentage. 1 WAD = funneling 100% to the burn engine
-    uint256     public           hop;  // [Seconds]   Time between kicks
-    uint256     public           zzz;  // [Timestamp] Last kick
+    uint256     public           live;    // Active Flag
+    FlapLike    public           flapper; // underlying burner strategy
```
oldchili, line 41:
```suggestion
    FlapLike    public           flapper; // Underlying burner strategy
```

oldchili (approved):
LGTM

sunbreak1211 (approved):
LGTM

PR #22, opened 2024-04-29
Upgradeable farm

sunbreak1211:


deploy/FlapperInit.sol
```diff
@@ -143,22 +150,41 @@ library FlapperInit {
         dss.chainlog.setAddress(clKey, wrapper_);
     }
 
+    function initFarm(
```
oldchili, line 153: The intention for this is to run on top of another initiation function for the farm?
Otherwise we might need to check more stuff like the staking token.
So maybe it should be mentioned that this is not really a regular init function (we also don't have a deploy function for the farm).

deploy/FlapperInit.sol
```diff
@@ -143,22 +150,41 @@ library FlapperInit {
...
+        address            farm_,
+        FarmConfig  memory cfg
+    ) internal {
+        FarmLike     farm     = FarmLike(farm_);
+        SplitterLike splitter = SplitterLike(cfg.splitter);
+
+        require(farm.rewardsToken() == DaiJoinLike(cfg.daiJoin).dai(), "Farm rewards not dai");
+
+        require(cfg.hop >= 5 minutes, "hop too low");
+        require(cfg.hop == splitter.hop(), "hop mismatch");
```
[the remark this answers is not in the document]
sunbreak1211: Not really assumes, it enforces it, otherwise the splitter hop would be 0, that's why added the >= 5 minutes as well, to avoid having 0 == 0 due to the splitter not being init before. But yeah adding a comment doesn't hurt.

deploy/FlapperInit.sol
sunbreak1211: Hmm I think that should be part of the init of Lockstake (check that every farm that is added, the staking token is `lsmkr`). In this side we only care about the rewards part as we have contact.
I agree it is a weird case and it goes out of the standard. For this specific case what I picture is you use the deploy script from the endgame-toolkit repo itself (or not, it doesn't really matter, as it is part of the spell review process to check a new contract is relied to the pause proxy and denied from the deployer), then the initiation happens in this repo init script verifying the rewards token is the one that the splitter will be sending, and the Lockstake makes sure that the farm/s being added are using the corresponding staking token.

sunbreak1211: Yeah ok, was referring more in a general sense.

deploy/FlapperInit.sol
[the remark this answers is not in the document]
oldchili: Ah ok, l didn't see that 5 minutes check as didn't pull today.
Let's still add a comment as you said.

deploy/FlapperInit.sol
```diff
@@ -143,22 +150,41 @@ library FlapperInit {
...
 
+    function initFarm(
+        DssInstance memory dss,
+        address            farm_,
+        FarmConfig  memory cfg
+    ) internal {
+        FarmLike     farm     = FarmLike(farm_);
+        SplitterLike splitter = SplitterLike(cfg.splitter);
+
+        require(farm.rewardsToken() == DaiJoinLike(cfg.daiJoin).dai(), "Farm rewards not dai");
```
oldchili, line 161: we should probably fetch `daiJoin` from the splitter and remove it from the `cfg`.

oldchili: And I guess this raises the question whether we also assume the splitter must also be init before the flapper, since in that case we can do the same in `initFlapperUniV2`. Maybe that for that we can leave it as is, since we don't gain that much by enforcing the order.

src/Splitter.sol
```diff
@@ -39,13 +39,13 @@ contract Splitter {
     mapping (address => uint256) public wards;
     uint256     public           live;    // Active Flag
     FlapLike    public           flapper; // Underlying burner strategy
+    FarmLike    public           farm;    // Underlying farming strategy
```
telome, line 42:
```suggestion
    FarmLike    public           farm;    // StakingRewards contract receiving DAI rewards
```

deploy/FlapperInit.sol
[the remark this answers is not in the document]
sunbreak1211: Comment added.

deploy/FlapperInit.sol
sunbreak1211: Hmm yeah not sure what to do here, not really strong opinion, but the lack of symmetry will look weird.

oldchili: ok, I don't feel strong about it, can leave as is.

sunbreak1211: Ok good let's leave this way then

oldchili (approved):
LGTM

test/Splitter.t.sol
```diff
@@ -157,14 +155,24 @@ contract SplitterTest is DssTest {
...
 
         DssInstance memory dss = MCD.loadFromChainlog(LOG);
         FlapperInit.initSplitter(dss, splitterInstance, splitterCfg);
         FlapperInit.initFlapperUniV2(dss, address(flapper), flapperCfg);
         FlapperInit.initDirectOracle(address(flapper));
+        FlapperInit.initFarm(dss, address(farm), farmCfg);
         vm.stopPrank();
 
         assertEq(dss.chainlog.getAddress("MCD_FLAP_SPLIT"), splitterInstance.splitter);
+        assertEq(dss.chainlog.getAddress("MCD_FLAP_BURN"), address(flapper));
```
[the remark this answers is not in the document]
telome: Ok, I'm fine to keep it as-is as well.

telome (approved):
LGTM

telome (approved):

oldchili (approved):
LGTM

PR #23, opened 2024-05-09
Readme Addition and Minor Comment Change

oldchili:


telome (approved):

src/Splitter.sol
```diff
@@ -39,7 +39,7 @@ contract Splitter {
     mapping (address => uint256) public wards;
     uint256     public           live;    // Active Flag
     FlapLike    public           flapper; // Underlying burner strategy
-    FarmLike    public           farm;    // StakingRewards contract receiving DAI rewards
+    FarmLike    public           farm;    // StakingRewards contract receiving the rewards
```
sunbreak1211, line 42: Not opposed to do the changed, but is there anything wrong with "DAI"?

oldchili: Capitalized DAI is assumed as the dai token.

sunbreak1211: Hmm yeah the whole thing is based on dai/DAI today, so this will need to be migrated, but yeah better to have one less place to change it.

sunbreak1211 (approved):
Looks good

src/Splitter.sol
oldchili: The migration should only be searching for `nst`, `Nst` and `NST`.
Something like `daiJoin` in the contract should be a generalized name for DAI's join and NST's join.

PR #25, opened 2024-07-22
Replace dai by nst

sunbreak1211:


README.md
```diff
@@ -4,17 +4,17 @@ Implementations of MakerDAO surplus auctions, triggered on `vow.flap`.
 
 ### Splitter
 
-Exposes a `kick` operation to be triggered periodically. Its logic withdraws `DAI` from the `vow` and splits it in two parts. The first part (`burn`) is sent to the underlying `flapper` contract to be processed by the burn engine. The second part (`WAD - burn`) is distributed as reward to a `farm` contract. The `kick` cadence is determined by the `hop` value.
+Exposes a `kick` operation to be triggered periodically. Its logic withdraws `NST` from the `vow` and splits it in two parts. The first part (`burn`) is sent to the underlying `flapper` contract to be processed by the burn engine. The second part (`WAD - burn`) is distributed as reward to a `farm` contract. The `kick` cadence is determined by the `hop` value.
```
telome, line 7: Should we use `vat.dai` here? 
```suggestion
Exposes a `kick` operation to be triggered periodically. Its logic withdraws `vat.dai` from the `vow`, converts it into `NST`, and splits it in two parts. The first part (`burn`) is sent to the underlying `flapper` contract to be processed by the burn engine. The second part (`WAD - burn`) is distributed as reward to a `farm` contract. The `kick` cadence is determined by the `hop` value.
```

telome: Maybe better to keep it simple, as you did.

sunbreak1211: Both seem correct to me. As it is true that initiates with a `vat.dai` move, but then exists to `NST` (as before it did to `DAI`). So I would probably slightly tend to leave it with `NST`.

deploy/FlapperInit.sol
```diff
@@ -112,14 +112,14 @@ library FlapperInit {
...
         // Sanity checks
         require(flapper.spotter()  == address(dss.spotter),                       "Flapper spotter mismatch");
-        require(flapper.dai()      == cfg.dai,                                    "Flapper dai mismatch");
+        require(flapper.nst()      == cfg.nst,                                    "Flapper nst mismatch");
         require(flapper.pair()     == cfg.pair,                                   "Flapper pair mismatch");
         require(flapper.receiver() == dss.chainlog.getAddress("MCD_PAUSE_PROXY"), "Flapper receiver mismatch");
 
         PairLike pair = PairLike(flapper.pair());
-        (address pairDai, address pairGem) = pair.token0() == cfg.dai ? (pair.token0(), pair.token1())
+        (address pairDai, address pairGem) = pair.token0() == cfg.nst ? (pair.token0(), pair.token1())
```
telome, line 120: Some local vars, types and functions were not renamed:
- in `src`: `pairDai`, `reserveDai`, `DaiJoinLike` (in the Splitter).
- in `test`: `reserveDai`, `minimalDaiReserve`, `currentDaiForGem`, `UNIV2_LINK_DAI_PAIR`, `_getDaiToSell`, `uniV2GemForDai`, `uniV2DaiForGem`, and some test function names. 

Was that intentional to reduce the diff? I'd tend to at least fully do the renaming for the files in `src`.

deploy/FlapperInit.sol
```diff
@@ -50,14 +50,14 @@ interface PairLike {
     function token1() external view returns (address);
 }
 
-interface DaiJoinLike {
-    function dai() external view returns (address);
+interface NstJoinLike {
```
telome, line 53: For consistency we should also rename `DaiJoinLike` > `NstJoinLike` in the Splitter (and SplitterMock).

README.md
telome: Ok, let's keep it as-is, resolving.

telome (approved):
LGTM

oldchili (approved):
LGTM

PR #31, opened 2024-09-19
Migrate tests to use USDS and SKY

sunbreak1211:


test/FlapperUniV2.t.sol
```diff
@@ -62,6 +62,10 @@ interface GemLike {
     function transfer(address, uint256) external;
 }
 
+interface Univ2FactoryLike {
```
telome, line 65: nit: we usually capitalize it as `UniV2XXX` in the rest of this repo.

test/Splitter.t.sol
```diff
@@ -141,23 +140,23 @@ contract SplitterTest is DssTest {
...
+            pair:            UNIV2_SKY_USDS_PAIR,
+            usds:            USDS,
             splitter:        address(splitter),
             prevChainlogKey: bytes32(0),
             chainlogKey:     "MCD_FLAP_BURN"
         });
         FarmConfig memory farmCfg = FarmConfig({
             splitter:        address(splitter),
-            usdsJoin:        DAI_JOIN,
+            usdsJoin:        USDS_JOIN,
```
telome, line 159: I think we can also now change `FlapperInit.setFarm` to use the `usds()` view function instead of `dai()` (and remove the corresponding TODO).

test/FlapperUniV2.t.sol
```diff
@@ -85,55 +89,62 @@ contract FlapperUniV2Test is DssTest {
...
-        MKR           = ChainlogLike(LOG).getAddress("MCD_GOV");
+        USDS          = ChainlogLike(LOG).getAddress("USDS");
+        SKY           = ChainlogLike(LOG).getAddress("SKY");
         USDC          = ChainlogLike(LOG).getAddress("USDC");
-        LINK          = ChainlogLike(LOG).getAddress("LINK");
         PAUSE_PROXY   = ChainlogLike(LOG).getAddress("MCD_PAUSE_PROXY");
         vat           = VatLike(ChainlogLike(LOG).getAddress("MCD_VAT"));
         vow           = VowLike(ChainlogLike(LOG).getAddress("MCD_VOW"));
+
+        UNIV2_USDS_IMX_PAIR = Univ2FactoryLike(UNIV2_FACTORY).createPair(USDS, IMX);
```
telome, line 126: We could maybe add a check that the pair doesn't already exist to make sure that the pair creation won't start failing at some point in the future.

test/FlapperUniV2.t.sol
```diff
@@ -310,89 +321,89 @@ contract FlapperUniV2Test is DssTest {
...
+        uint256 totalUsdsConsumed = initialUsds - GemLike(USDS).balanceOf(address(this));
+        uint256 boughtLpOldVersion = GemLike(UNIV2_SKY_USDS_PAIR).balanceOf(PAUSE_PROXY) - initialLp;
 
         vm.revertTo(initialState);
 
         // New version
-        vm.prank(PAUSE_PROXY); vow.file("bump", totalDaiConsumed * RAY); // The current flapper gets the total vat.dai to consume.
-        doExec(address(flapper), MKR, UNIV2_DAI_MKR_PAIR);
-        uint256 boughtLpNewVersion = GemLike(UNIV2_DAI_MKR_PAIR).balanceOf(PAUSE_PROXY) - initialLp;
+        vm.prank(PAUSE_PROXY); vow.file("bump", totalUsdsConsumed * RAY); // The current flapper gets the total vat.Usds to consume.
```
telome, line 402:
```suggestion
        vm.prank(PAUSE_PROXY); vow.file("bump", totalUsdsConsumed * RAY); // The current flapper gets the total vat.dai to consume.
```

test/Splitter.t.sol
```diff
@@ -175,17 +174,17 @@ contract SplitterTest is DssTest {
         assertEq(dss.chainlog.getAddress("MCD_FARM_USDS"), address(farm));
 
         // Add initial liquidity if needed
-        (uint256 reserveDai, ) = UniswapV2Library.getReserves(UNIV2_FACTORY, DAI, MKR);
-        uint256 minimalDaiReserve = 280_000 * WAD;
-        if (reserveDai < minimalDaiReserve) {
+        (uint256 reserveUsds, ) = UniswapV2Library.getReserves(UNIV2_FACTORY, USDS, SKY);
+        uint256 minimalUsdsReserve = 280_000 * WAD;
+        if (reserveUsds < minimalUsdsReserve) {
```
telome, line 179: Maybe more clear if the next line uses a realistic SKY price (e.g. 0.06 ether)?

test/FlapperUniV2.t.sol
```diff
@@ -85,55 +89,62 @@ contract FlapperUniV2Test is DssTest {
...
+            deal(USDS, UNIV2_USDS_IMX_PAIR, 10_000_0000 * WAD, true);
+            PairLike(UNIV2_USDS_IMX_PAIR).sync();
+        }
+
+        (flapper, medianizer) = setUpFlapper(SKY, UNIV2_SKY_USDS_PAIR, 727 * WAD, "MCD_FLAP") ;
+        assertEq(flapper.usdsFirst(), false);
 
-        (linkFlapper, linkMedianizer) = setUpFlapper(LINK, UNIV2_LINK_DAI_PAIR, 654 * WAD / 100, bytes32(0));
-        assertEq(linkFlapper.usdsFirst(), false);
+        (imxFlapper, imxMedianizer) = setUpFlapper(IMX, UNIV2_USDS_IMX_PAIR, 654 * WAD / 100, bytes32(0));
```
telome, line 144: Same comment as above: we could use a more realistic price for SKY (and IMX) to avoid confusion. The same applies to the SwapOnly test file.

test/FlapperUniV2.t.sol
telome: And we could move the code adding the initial liquidity (L135-139) under the same `if`.

test/Splitter.t.sol
sunbreak1211: Done.

test/FlapperUniV2.t.sol
sunbreak1211: Done.

test/FlapperUniV2.t.sol
sunbreak1211: Fixed

test/FlapperUniV2.t.sol
sunbreak1211: Done

test/Splitter.t.sol
sunbreak1211: Yes I was wondering what to do on prices as the base pip is the MKR one which is confusing. Maybe I should just use any other random as the only important thing is actually the contract functionality.

telome: Yes or if it's not too much change we could also use the OracleWrapper with the PIP_MKR.

test/FlapperUniV2.t.sol
```diff
@@ -85,55 +90,65 @@ contract FlapperUniV2Test is DssTest {
...
         USDC          = ChainlogLike(LOG).getAddress("USDC");
-        LINK          = ChainlogLike(LOG).getAddress("LINK");
         PAUSE_PROXY   = ChainlogLike(LOG).getAddress("MCD_PAUSE_PROXY");
         vat           = VatLike(ChainlogLike(LOG).getAddress("MCD_VAT"));
         vow           = VowLike(ChainlogLike(LOG).getAddress("MCD_VOW"));
-        
-        splitter = new SplitterMock(DAI_JOIN);
+
+        UNIV2_USDS_IMX_PAIR = UniV2FactoryLike(UNIV2_FACTORY).getPair(USDS, IMX);
+        if(UNIV2_USDS_IMX_PAIR == address(0)) {
```
telome, line 128:
```suggestion
        if (UNIV2_USDS_IMX_PAIR == address(0)) {
```

telome: Also applies to a few other places.

sunbreak1211: oops fixed in both places

test/Splitter.t.sol
sunbreak1211: Yeah at some point started to do that as well, but then I saw the flappers test use a mock and the oracle wrapper test PIP_ETH that wasn't convinced what path to take. Will give a bit more of thought to that to try to standardise a bit.

test/FlapperUniV2.t.sol
sunbreak1211: ahh ok, fixed as well!

test/FlapperUniV2.t.sol
telome: > And we could move the code adding the initial liquidity (L135-139) under the same `if`.

Not sure if you missed that comment above ^?

sunbreak1211: Ah yeah, I think I didn't have the second comment in the page when doing (or maybe just missed it). However wondering if we want to do that as I believe function that adapts the price assumes to have certain volume and even if the pair exists doesn't mean that has it. But will double check as might be confused

sunbreak1211: Ok yeah, whatever volume is fine, just 0 is the problem. Not sure what we prefer, as probably if created will receive some volume, but we might have the edge case that was created but nobody deposited and checking that seems overkilling.

telome: Ah yes you're right, better to leave it as you did I think. Resolving.

oldchili (approved):
LGTM

telome (approved):
Looks good

PR #34, opened 2026-04-21
Add Beam

sunbreak1211:


src/SBEBeam.sol
```diff
@@ -0,0 +1,163 @@
...
+        require(hop >= minHop,           "SBEBeam/hop-below-min");
+        require(hop < type(uint256).max, "SBEBeam/hop-halts-engine");
+        require(kbump / hop <= maxRate,  "SBEBeam/rate-above-max");
+
+        kicker.file("kbump", kbump);
+        splitter.file("burn", burn);
+        if (hop != splitter.hop()) {
+            // Avoid to extend duration of current stream if hop did not change
+            splitter.file("hop", hop);
+            farmOwner.setRewardsDuration(hop);
```
oldchili: I think there could be an option where a farm is not used/set.
Maybe we should make `"farmOwner" fileable?

src/SBEBeam.sol
```diff
@@ -0,0 +1,163 @@
...
+interface FarmOwnerLike {
+    function setRewardsDuration(uint256) external;
+}
+
+contract SBEBeam {
+    // --- storage variables ---
+
+    mapping(address => uint256) public wards;
+    mapping(address => uint256) public buds;
+    uint256 public maxKbump; // [rad]     Maximum allowed value for Kicker.kbump
```
oldchili, line 46: There is a corner case where kicks didn't happen for a long time, or maybe never happened (if Splitter was just deployed). In that case `splitter.zzz` will be far in the past, so a kick could happen even if `hop` is very large. It also means a large `kbump` could be configured by the bud along with this large `hop`, and could still satisfy `maxRate`.

Therefore in terms of security assumptions I think we should assume a first kick of size `maxKbump` would always be possible, and this emphasizes that `maxKbump` should always be used and thought of.

I suggest to add a comment/warning about its importance.

src/SBEBeam.sol
```diff
@@ -0,0 +1,163 @@
...
+        require(kbump <= maxKbump,       "SBEBeam/kbump-above-max");
+        require(kbump % RAY == 0,        "SBEBeam/kbump-not-multiple-of-RAY");
+        require(burn <= WAD,             "SBEBeam/burn-above-max");
+        require(hop >= minHop,           "SBEBeam/hop-below-min");
+        require(hop < type(uint256).max, "SBEBeam/hop-halts-engine");
+        require(kbump / hop <= maxRate,  "SBEBeam/rate-above-max");
+
+        kicker.file("kbump", kbump);
+        splitter.file("burn", burn);
+        if (hop != splitter.hop()) {
```
oldchili: Here we assume that `splitter.hop` and `farm.rewardsDuration` never desync, as we only set the farm reward duration looking at `splitter.hop`.
I suggest to either compare to both, or add a comment about this assumption.

src/SBEBeam.sol
```diff
@@ -0,0 +1,163 @@
...
+    // - It is intended to rewrite the same values, emit the event, and reset the toc count, even if there is no change.
+    // - Only the throughput-increasing directions are bounded: kbump is capped at maxKbump, hop is
+    //   floored at minHop, and the burn rate (kbump / hop) is capped at maxRate. Lowering kbump or
+    //   raising hop is always allowed; at worst it stalls the burn stream, which governance can revive.
+    // - burn is capped at WAD (100%); a higher value would make Splitter.kick underflow and halt.
+    // - kbump must be a whole multiple of RAY, preserving the Kicker deploy invariant and avoiding kick dust.
+    // - hop must stay below type(uint256).max: that value is the halt sentinel (see the good modifier),
+    //   reserved for governance, so a facilitator cannot use set() to halt and lock itself out of the module.
+    function set(uint256 kbump, uint256 burn, uint256 hop) external toll good {
+        require(block.timestamp >= tau + toc, "SBEBeam/too-early");
```
oldchili: In general since we removed `step` it is not super-obvious that we even need beam rate limiting (`toc` and `tau`) anymore.

I guess we can leave it to counter any unforeseen farm-related corner cases due to calling `setDuration` very often (i.e defense in depth).

src/SBEBeam.sol
```diff
@@ -0,0 +1,163 @@
...
+    function diss(address usr) external auth {
+        buds[usr] = 0;
+        emit Diss(usr);
+    }
+
+    function file(bytes32 what, uint256 data) external auth {
+        if (what == "maxKbump") {
+            maxKbump = data;
+        } else if (what == "minHop") {
+            minHop = data;
```
oldchili, line 123: Should we require `cfg.minHop >= 5 minutes` as done in `initSplitter`?

deploy/FlapperInit.sol
```diff
@@ -247,4 +274,49 @@ library FlapperInit {
...
+        address            farmOwner
+    ) internal {
+        address splitter = dss.chainlog.getAddress("MCD_SPLIT");
+        address farm     = SplitterLike(splitter).farm();
+
+        require(FarmOwnerLike(farmOwner).farm() == farm, "FarmOwner farm mismatch");
+
+        FarmLike(farm).nominateNewOwner(farmOwner);
+        FarmOwnerLike(farmOwner).acceptOwnership();
+    }
```
oldchili, line 300: should we add `farmOwner` to the chainlog?

deploy/FlapperInit.sol
```diff
@@ -68,16 +69,33 @@ interface FarmLike {
     function rewardsToken() external view returns (address);
     function setRewardsDistribution(address) external;
     function setRewardsDuration(uint256) external;
+    function nominateNewOwner(address) external;
+}
+
+interface FarmOwnerLike {
+    function farm() external view returns (address);
+    function wards(address) external view returns (uint256);
```
oldchili:
```suggestion
```

Unused

deploy/FlapperInit.sol
```diff
@@ -247,4 +274,49 @@ library FlapperInit {
...
+        FarmLike(farm).nominateNewOwner(farmOwner);
+        FarmOwnerLike(farmOwner).acceptOwnership();
+    }
+
+    function initSBEBeam(
+        DssInstance   memory dss,
+        address              beam,
+        address              farmOwner,
+        SBEBeamConfig memory cfg
+    ) internal {
```
oldchili, line 307: In general we assume here that there is a farm configured (since we call into `farmOwner`), but it could be in theory that no farm is used and we are planning to use the beam just for burns.

I think we should at least state that we assume that farm is in use and are aware of the above option.

deploy/FlapperInit.sol
```diff
@@ -247,4 +274,49 @@ library FlapperInit {
...
+        address kicker     = dss.chainlog.getAddress("MCD_KICK");
+        address splitter   = dss.chainlog.getAddress("MCD_SPLIT");
+
+        // Sanity checks
+        require(SBEBeamLike(beam).kicker()    == kicker,    "SBEBeam kicker mismatch");
+        require(SBEBeamLike(beam).splitter()  == splitter,  "SBEBeam splitter mismatch");
+        require(SBEBeamLike(beam).farmOwner() == farmOwner, "SBEBeam farmOwner mismatch");
+
```
oldchili, line 316: Should we add something like - `require(FarmLike(splitter.farm()).owner() == farmOwner`, "SBEBeam/farm-not-owned");

That would make sure that `initFarmOwner` was called.
This will also add an ordering constraint, which I find healthy (more deterministic, less code paths).

src/SBEBeam.sol
```diff
@@ -0,0 +1,163 @@
...
+    // Notes:
+    // - It is intended to rewrite the same values, emit the event, and reset the toc count, even if there is no change.
+    // - Only the throughput-increasing directions are bounded: kbump is capped at maxKbump, hop is
+    //   floored at minHop, and the burn rate (kbump / hop) is capped at maxRate. Lowering kbump or
+    //   raising hop is always allowed; at worst it stalls the burn stream, which governance can revive.
+    // - burn is capped at WAD (100%); a higher value would make Splitter.kick underflow and halt.
+    // - kbump must be a whole multiple of RAY, preserving the Kicker deploy invariant and avoiding kick dust.
+    // - hop must stay below type(uint256).max: that value is the halt sentinel (see the good modifier),
+    //   reserved for governance, so a facilitator cannot use set() to halt and lock itself out of the module.
+    function set(uint256 kbump, uint256 burn, uint256 hop) external toll good {
```
oldchili: There could be questions on why we don't support setting `kicker.khump`.
Suggest to add a comment about it somewhere.

src/SBEBeam.sol
```diff
@@ -0,0 +1,163 @@
...
+
+contract SBEBeam {
+    // --- storage variables ---
+
+    mapping(address => uint256) public wards;
+    mapping(address => uint256) public buds;
+    uint256 public maxKbump; // [rad]     Maximum allowed value for Kicker.kbump
+    uint256 public minHop;   // [seconds] Minimum allowed value for Splitter.hop (also applied to farm.rewardsDuration)
+    uint256 public maxRate;  // [rad/s]   Maximum allowed burn rate (kbump / hop)
+    uint64  public tau;      // Cooldown period between set() calls in seconds
```
oldchili, line 49: Just a note that in the splitter these rate limiting vars are uint256, but I see the current implementations aligns with sp-beam so I think it's fine.

README.md
```diff
@@ -52,6 +52,43 @@ Configurable Parameters:
 
 This contract allows bypassing the governance delay when disabling the Splitter in an emergency.
 
+### FarmOwner
```
oldchili, line 55: A bit too much documentation to my personal taste.
For example not sure we need to detail all the forwarded messages, or we need the `Access control` part.

src/SBEBeam.sol
oldchili, line 46: Just a note that the `bud` (semi-trusted) can steal funds that are distributed while its in control as it in theory can set `burn` to 0% and stake a large portion of the farm (although would not be easy to acquire such amounts of SKY atm).

OFC `maxKbump`, `minHop` and `maxRate` are designed to protect against this.

src/SBEBeam.sol
```diff
@@ -0,0 +1,163 @@
...
+            maxRate = data;
+        } else if (what == "tau") {
+            require(data <= type(uint64).max, "SBEBeam/invalid-tau-value");
+            tau = uint64(data);
+        } else if (what == "toc") {
+            require(data <= type(uint128).max, "SBEBeam/invalid-toc-value");
+            toc = uint128(data);
+        } else revert("SBEBeam/file-unrecognized-param");
+        emit File(what, data);
+    }
```
oldchili, line 134: Note:
Since the farm may be used in LSE liquidations we need to make sure the bud can not freeze withdrawals from it, even temporarily.

The farm's `withdraw` is gated solely by the `updateReward` modifier in `rewardPerToken()`, so the only way to freeze it would be to push `rewardRate` high enough to overflow.

I didn't finds a way this could happen with the current farm code, buy worth looking into more.

src/FarmOwner.sol
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    function deny(address usr) external auth {
+        wards[usr] = 0;
+        emit Deny(usr);
+    }
+
+    // --- forwarded onlyOwner methods + acceptOwnership ---
+
+    function setRewardsDuration(uint256 duration) external auth {
```
oldchili, line 81: In theory it's not necessary to let the beam call anything apart from `setRewardsDuration`, so giving it full auth in this contract is not a must (and could be problematic if the beam code for some reason is changed to be able to call more functions).

I guess we can consider it a design choice, but worth the discussion.

src/SBEBeam.sol
```diff
@@ -0,0 +1,163 @@
...
+}
+
+contract SBEBeam {
+    // --- storage variables ---
+
+    mapping(address => uint256) public wards;
+    mapping(address => uint256) public buds;
+    uint256 public maxKbump; // [rad]     Maximum allowed value for Kicker.kbump
+    uint256 public minHop;   // [seconds] Minimum allowed value for Splitter.hop (also applied to farm.rewardsDuration)
+    uint256 public maxRate;  // [rad/s]   Maximum allowed burn rate (kbump / hop)
```
oldchili: Note:
Since it could be that `now > zzz + hop` for a given moment, it might be possible to do 2 kicks over one `hop` (one in the beginning and one right in the end).

So it could be argued that `maxRate` is not an exact rate.
I think it is fine and mostly writing it here so it's clear to be taken into consideration.

test/Kicker.t.sol
```diff
@@ -125,6 +125,9 @@ contract KickerTest is DssTest {
 
         // Allow Test contract to read from Scribe oracle
         vm.store(address(medianizer), keccak256(abi.encode(address(this), uint256(2))), bytes32(uint256(1)));
+
+        // Add surplus
+        vm.store(address(dss.vat), keccak256(abi.encode(address(dss.vow), uint256(5))), bytes32(uint256(dss.vat.sin(address(dss.vow)) + 50_000_000e45)));
```
oldchili, line 130: Would be a bit nicer with checked write:

`stdstore.target(address(dss.vat)).sig("dai(address)").with_key(address(dss.vow)).depth(0).checked_write(dss.vat.sin(address(dss.vow)) + 50_000_000e45)`

test/SBEBeam.t.sol
```diff
@@ -0,0 +1,348 @@
...
+        vm.expectRevert("SBEBeam/hop-halts-engine");
+        vm.prank(bud);
+        beam.set(5_000e45, 0.5e18, type(uint256).max);
+    }
+
+    // The throttling-only directions are never blocked: kbump can be lowered arbitrarily
+    // and hop raised arbitrarily, since at worst that stalls the burn stream.
+    function testSetKbumpCanGoArbitrarilyLowAndHopCanGoArbitrarilyHigh() public {
+        vm.prank(bud);
+        beam.set(1e27, 0.5e18, 365 days); // far below the previous kbump (still a RAY multiple), far above the previous hop
```
oldchili: `1e27` is a bit confusing here, I'd expect either `1e45` or maybe just `1`.
`e27` resolution is not really used here.

test/SBEBeam.t.sol
```diff
@@ -0,0 +1,348 @@
...
+        assertEq(kicker.kbump(),         1e27);
+        assertEq(splitter.hop(),         365 days);
+        assertEq(farm.rewardsDuration(), 365 days);
+    }
+
+    function testSetAboveMaxBurn() public {
+        vm.expectRevert("SBEBeam/burn-above-max");
+        vm.prank(bud);
+        beam.set(5_000e45, WAD + 1, 1 hours); // burn above WAD (100%) would halt Splitter.kick
```
oldchili:
```suggestion
        beam.set(5_000e45, WAD + 1, 1 hours); // burn above WAD (100%) would revert
```

test/SBEBeam.t.sol
```diff
@@ -0,0 +1,348 @@
...
+    function testSetRateAtMax() public {
+        vm.prank(pauseProxy);
+        beam.file("maxRate", 1e45); // 3_600e45 / 1 hours
+
+        vm.prank(bud);
+        beam.set(3_600e45, 0.5e18, 1 hours); // 3_600e45 / 3600 = 1e45 == maxRate
+        assertEq(kicker.kbump(), 3_600e45);
+        assertEq(splitter.hop(), 1 hours);
+    }
+}
```
oldchili, line 582: Maybe worth adding a test that even a very small hop can not freeze farm withdrawals (and thus potentially LSE liquidations) ?

src/SBEBeam.sol
sunbreak1211: So yeah, calling multiple times the `setRewardsDuration` will make that the existing funds of the farm keep stretching out in time, so it is a way of DoS which I think is the only reason to leave the time delay between calls.
On the other hand, we are already assuming that the beam can provoke DoS not sending new funds (different specific thing but same nature), then we might opt to just simplify and remove the time delay between calls and let governance to fix a situation like that.

src/SBEBeam.sol
sunbreak1211: Yeah, good idea, wondering if we should add it here or in the `set` call. As there would be more homogeneous with the % RAY check and also 100% enforced.

deploy/FlapperInit.sol
sunbreak1211: Yeah was wondering that. I wasn't sure it had enough importance to be, an also annoying that we add to the chainlog, but then we can not use it from the chainlog in the `initSBEBeam` because the way of this init script where all the keys are configurable. I guess the only option is passing the key to query instead of the address.

oldchili: I think we can add `farmOwner` to the chainlog and keep on passing the address to the other function as now.

src/SBEBeam.sol
oldchili: The DoS is not that concerning as it is temporary.
Not sure about removing it altogether though.

src/SBEBeam.sol
oldchili: Would be a bit weird/missleading to allow a `minHop` that is lower than it though.
That's why it might be best to have it in the `file`.

deploy/FlapperInit.sol
sunbreak1211: Done

deploy/FlapperInit.sol
sunbreak1211: Done

src/SBEBeam.sol
sunbreak1211: "`Kicker.khump` (the flap threshold) is deliberately left out of the `set` knobs: it is not a value that needs regular tuning, and changing it is a more structural governance decision better routed through the full governance process."

What do you think?

src/SBEBeam.sol
sunbreak1211: you mean about the tau and toc?

README.md
sunbreak1211: Removed the forwarded messages section

src/SBEBeam.sol
sunbreak1211: ok yeah, you want to add a comment about that?

src/FarmOwner.sol
sunbreak1211: I'm not sure it is worth to make FarmOwner more complex when this model of relying the whole thing by a contract that has very limited access already exists in other places. But not fully opposed if you feel is worth.

src/SBEBeam.sol
sunbreak1211: do you think it is worth to mention that? how would you frame in that side comment?

test/SBEBeam.t.sol
sunbreak1211: It is the minimum accepted value due to `require(kbump % RAY == 0,        "SBEBeam/kbump-not-multiple-of-RAY");`

test/Kicker.t.sol
sunbreak1211: Well, it uses the exact same format than the one line above storage change, and it is about something not related to the current PR so I would prefer to leave it with the least possible of changes.

test/SBEBeam.t.sol
sunbreak1211: Hmmm but what exactly do you want to test here? a specific farm behavior? because if that is the case then we probably need to remove the mock and import the exact dependency or maybe we should be using the deployed farm.

src/SBEBeam.sol
sunbreak1211: Hmm the farm is already deployed and set up, not sure if worth making this a bit more complex for a case of a different deployment. In the worst case that there is full burn we would be updating the farm without any purpose but won't be harmful either.

src/SBEBeam.sol
sunbreak1211: Hmmm yeah, this seems to be an invariant that exceeds this case, but I guess we can add a comment.

sunbreak1211: If you prefer to check both we could also do, and then the beam would also be a way to fix a possible external desync

src/SBEBeam.sol
sunbreak1211: Well not sending new funds is not much different than making the existing too spared in time. Or maybe you are referring to something else.

test/SBEBeam.t.sol
sunbreak1211: Changed

src/SBEBeam.sol
sunbreak1211: So we could add something like this at the end of SBEBeam section:
"
There is a corner case where kicks haven't happened for a long time, or they might not even existed at all (if `Splitter` was just deployed). In that case `Splitter.zzz` would be far in the past, so a `kick` could execute immediately (or very soon) even if `hop` is set large. That means a large `kbump` could be configured by the `bud` along with this large `hop` satisfying what was thought a safe `maxRate`.
Therefore in terms of security assumptions, it has to be assumed a first `kick` of size `maxKbump` would be potentially possible, then then `maxKbump` should always be defined as a safe standalone value.
"

src/SBEBeam.sol
sunbreak1211: Or are you thinking there might be another issue?

src/SBEBeam.sol
sunbreak1211: Done

deploy/FlapperInit.sol
sunbreak1211: Done

src/SBEBeam.sol
sunbreak1211: Hmmm, it is true that the farm is upgradable in the Splitter, so maybe it makes sense to do the `farmOnwer` also upgradable in the Beam mainly to keep an unified criteria.

src/SBEBeam.sol
oldchili: Yes, maybe a slightly shorter and maybe more generic version:

"
Note that `maxKbump` must be safe as a standalone kick size, not just safe relative to `hop` or `maxRate`. Because hop only gates the gap since the last kick (`Splitter.zzz`), a long-idle or freshly-deployed Splitter lets a kick fire immediately regardless of `hop`.
"

src/SBEBeam.sol
oldchili: Yeah, I think we can set each one separately and add some comment that they are supposed to be the same, but if not this implementation also solves a possible desync for whatever reason.

The code (without the comment) would then look something like this:

```
  if (hop != splitter.hop()) {  
      splitter.file("hop", hop);
  }

  if (hop != farm.rewardsDuration()) {
      farmOwner.setRewardsDuration(hop);
  }
```

src/SBEBeam.sol
oldchili: let's maybe leave this one for the end of the review

src/SBEBeam.sol
oldchili: Sounds good.

src/SBEBeam.sol
oldchili: Yes:

```
  uint64  public tau;   // cooldown duration
  uint128 public toc;   // last set() timestamp
```

src/SBEBeam.sol
oldchili: I think it's fine to not have a special comment as we didn't really say otherwise.
Was just a note in case we need to show someone this was known.
Resolving.

src/FarmOwner.sol
oldchili: Ok, I'm fine with that.
Resolving.

src/SBEBeam.sol
oldchili: I guess it's fine, and we can point anyone here if needed.
Resolving.

test/Kicker.t.sol
oldchili: Sounds good, resolving.

test/SBEBeam.t.sol
oldchili: Got it.

test/SBEBeam.t.sol
oldchili: Yeah I meant the currently used LSSKY=>USDS farm.
Maybe this is just something to test loca.ly and we don't have to commit.

sunbreak1211: Let me think a bit if we can do somehow, but yeah, we could do that otherwise

src/SBEBeam.sol
sunbreak1211: Added! (with minimal changes)

src/SBEBeam.sol
sunbreak1211: Added

src/SBEBeam.sol
sunbreak1211: ok yeah, I think it is just find as it is as well

src/SBEBeam.sol
sunbreak1211: Done

src/SBEBeam.sol
sunbreak1211: Implemented the change. Made also not to call the `setRewardsDuration` when the `burn` equals to `WAD`.
That way we could allow not having a farm set.
The only thing I'm wondering if we should also have a sanity check in `set` that `splitter.farm == farmOwner.farm`. I would vote for adding it.

deploy/FlapperInit.sol
sunbreak1211: It allows now to run without a farmOwner set (in the meanwhile `burn == WAD`)

test/SBEBeam.t.sol
sunbreak1211: Actually nvm, we are using the real farm here. The thing is I still find it weird to test the `hop` itself as now we have the 5 minutes protection, which is already something we have been using. What I might agree is actually trying to test a scenario where we call a lot of times `setRewardsDuration`.

src/SBEBeam.sol
oldchili: Yeah we can have the sanity check.

Also the current version looks good to me, just maybe add a line break before the long comment (which already hurts my eyes as is :)).

deploy/FlapperInit.sol
oldchili: My comment here was about the init script.
It would currently fail if farmOwner is addres(0), as it calls `rely` on it.

test/SBEBeam.t.sol
oldchili: yes, agree.

src/SBEBeam.sol
sunbreak1211: Done

oldchili: we can also add this sanity check prior, so we have a nicer revert if `farmOwner` is not set:
`require(burn == WAD || address(farmOwner) != address(0), "SBEBeam/farm-owner-required");`

src/SBEBeam.sol
```diff
@@ -0,0 +1,188 @@
...
+        splitter.file("hop", hop);
+
+        // When burn == WAD all surplus goes to the burn engine and nothing is
+        // staked, so there is no farm reward stream to re-rate. Otherwise:
+        // - avoid extending the duration of the current stream if hop did not change, and
+        // - indirectly allow fixing a possible desync between splitter and the
+        //   farm for whatever reason that could have happened (included a prior burn=WAD).
+        if (burn < WAD && hop != FarmLike(farmOwner.farm()).rewardsDuration()) {
+            farmOwner.setRewardsDuration(hop);
+        }
```
oldchili, line 181: There is a problem that a large enough hop sets the farm's reward rate to zero, and re-rating can't undo it.

The farm derives its reward rate as committed / hop. If a bud sets a hop larger than the committed reward, the rate floors to zero — and it gets stuck there:

setRewardsDuration recomputes the committed amount from the (now zero) rate, so re-rating back to a sane hop leaves the rate at zero. The Beam or governance can't restore the reward stream through set() — distribution just stops, and recovery requires a fresh reward notification (or a recoverERC20), not the normal re-rate path.

Test demonstrating it (passes on a mainnet fork):

```
  interface FarmRewardsLike {
      function rewardsToken() external view returns (address);
      function rewardsDistribution() external view returns (address);
      function rewardRate() external view returns (uint256);
      function getRewardForDuration() external view returns (uint256);
      function notifyRewardAmount(uint256) external;
  }
  
  interface ERC20Like {
      function balanceOf(address) external view returns (uint256);
  }

  function testSetHugeHopStrandsFarmRewards() public {
      FarmRewardsLike f = FarmRewardsLike(address(farm));
      ERC20Like usds = ERC20Like(f.rewardsToken());

      // Clean slate, then fund a normal 1h reward stream.
      vm.warp(block.timestamp + 365 days);
      uint256 reward = 3_600e18;
      deal(address(usds), address(farm), reward);
      vm.prank(f.rewardsDistribution());
      f.notifyRewardAmount(reward);

      assertGt(f.rewardRate(), 0);           // live stream
      assertGt(f.getRewardForDuration(), 0); // ~reward scheduled
      uint256 farmBal = usds.balanceOf(address(farm));

      // Bud sets a hop larger than the committed reward -> rewardRate floors to 0.
      uint256 hugeHop = 1e25; // >= minHop, < type(uint256).max
      vm.prank(bud);
      beam.set(5_000e45, 0.5e18, hugeHop);
      assertEq(f.rewardRate(), 0);           // rate wiped to zero

      // Re-rating back to a sane hop does NOT restore it.
      vm.prank(bud);
      beam.set(5_000e45, 0.5e18, 1 hours);
      assertEq(f.rewardRate(), 0);           // still 0 — re-rating can't undo it
      assertEq(f.getRewardForDuration(), 0); // nothing scheduled

      // Tokens remain in the farm but the stream is dead.
      assertEq(usds.balanceOf(address(farm)), farmBal);
  }
```
  
Given this, I think we might need a maxHop ceiling afterall. Bounding hop keeps rewardRate above zero so the stream is always recoverable via set().

src/SBEBeam.sol
sunbreak1211: Added a fuzzing test regarding this, let me know what you think

test/SBEBeam.t.sol
sunbreak1211: Check the test referred in the other comment

src/SBEBeam.sol
sunbreak1211: Looking your new comment though.

src/SBEBeam.sol
sunbreak1211: Added hard cap

src/SBEBeam.sol
sunbreak1211: I have embed the condition in the existing require, it won't have a separate revert message but I think is overall better than duplicating.

deploy/FlapperInit.sol
sunbreak1211: I put a comment in the init script, let me know if it is fine or you prefer we actually allow initializing without a farm (I personally don't think it is worth).

src/SBEBeam.sol
sunbreak1211: Nothing really substantial changed in this fuzzing test regarding that other comment

src/SBEBeam.sol
oldchili: Looks good.
I'd only maybe remove the  mentioning that `notifyRewardAmount` would revive the farm (in the code and readme), as that's true, but it would keep the prior zeroed funds unaccounted for.

In general Claude adds too much wording imo.

src/SBEBeam.sol
oldchili: Nice.
I'm not 100% sure that a fuzzing test is the right tool here, but I guess it's good enough.

I also don't really see a way to block withdrawals, so I think we can probably conclude this (and even if there is some very far fetched way, Sky liquidations are turned off with no enabling plan, plus the bud is semi-trusted, so I'd rank it as low severity).

Resolving.

test/SBEBeam.t.sol
oldchili: yes, resolved the other comment, doing the same here.

deploy/FlapperInit.sol
```diff
@@ -247,4 +274,54 @@ library FlapperInit {
...
+        SBEBeamConfig memory cfg
+    ) internal {
+        address kicker   = dss.chainlog.getAddress("MCD_KICK");
+        address splitter = dss.chainlog.getAddress("MCD_SPLIT");
+        address farm     = SplitterLike(splitter).farm(); // Assuming there will be a farm set when initializing SBEBeam
+
+        // Sanity checks
+        require(SBEBeamLike(beam).kicker()   == kicker,    "SBEBeam kicker mismatch");
+        require(SBEBeamLike(beam).splitter() == splitter,  "SBEBeam splitter mismatch");
+        require(FarmLike(farm).owner()       == farmOwner, "SBEBeam farm not owned");
```
telome:
```suggestion
        require(FarmLike(farm).owner()       == farmOwner, "Farm owner mismatch");
```

src/SBEBeam.sol
sunbreak1211: Well I guess the `pauseProxy` can call `setRewardsDistribution` to itself, then call `notifyRewardAmount` (with the balance of the contract), and then set back the vested distribution back. But fine if we still want to remove that.

oldchili: Yeah, I'd not go into it.

deploy/FlapperInit.sol
sunbreak1211: Changed

src/SBEBeam.sol
sunbreak1211: Made a change in the README to make it less specific

sunbreak1211: let me know

telome: The Note in the code still has the old comment. 

Not sure it's technically correct that "no further set() can revive it" as I think a `bud` could fix the `hop` with a `set()` and then we'd wait for another `kick()` to call `notifyRewardAmount` (assuming `burn < WAD`) for the `rewardRate` to be fixed. Though as I understand it you'll need governance to intervene to recover the unaccounted funds.

deploy/FlapperInit.sol
telome: nit: this init file seems to be using the format "\<contractName\> \<getterName\> mismatch" (though not everywhere), so "Farm owner mismatch" might be slightly more appropriate.

src/SBEBeam.sol
sunbreak1211: Change that one and a bit more of the README, let me know what you think

deploy/FlapperInit.sol
sunbreak1211: Ah yes, hadn't realized, you are right, fixed.

src/SBEBeam.sol
telome: Sounds good to me.

oldchili: Looks good, resolving.

oldchili (dismissed):
Looks good to me.
I'm also fine with removing `tau` and `toc` as discussed in the thread.
Will re-approve if needed.

src/SBEBeam.sol
telome: Might be worth also pointing out that the initial rate might exceed `maxRate` due to this possibility of kicking `maxKbump + maxRate * W` over some initial window `W`.  So the effective rate could reach `maxRate + maxKbump/W > maxRate`.

oldchili: Yeah, I think this is also discussed here - https://github.com/sky-ecosystem/dss-flappers/pull/34#discussion_r3450017617 .

telome: Yeah, saw that, though I think that in the case of the Splitter being idle for some time you can attribute the first reviving kick to the period before it, so `maxRate` is still arguably respected. But you can't do that with the very first kick of a freshly deployed splitter. 

But in any case, I agree that `maxRate` should be thought of as an average rate over a sufficiently long period of time, not something to be considered enforced over short windows.

src/SBEBeam.sol
telome: Haven't found any issue other than the DoS you mentioned that the tau check really mitigates, so fine by me to remove that check (or to keep it as defense-in-depth).

sunbreak1211: Ok, we have decided to leave it. I guess this can be resolved.

oldchili: Yes, even if not strictly necessary it also gives users some assurance that params would not change too often. Resolving.

test/SBEBeam.t.sol
```diff
@@ -0,0 +1,573 @@
...
+        emit Set(kbump_, burn_, hop_);
+        vm.prank(bud);
+        beam.set(kbump_, burn_, hop_);
+
+        assertEq(beam.toc(), block.timestamp);
+    }
+
+    function testSetSameHopSkipsRewardsDuration() public {
+        // hop is left unchanged, so the farm's reward stream must not be re-rated:
+        // neither splitter.file("hop", ...) nor farm.setRewardsDuration(...) should be called.
```
telome:
```suggestion
        // farm.setRewardsDuration(...) should not be called.
```

telome (dismissed):
Looks good

deploy/FlapperInit.sol
```diff
@@ -247,4 +274,54 @@ library FlapperInit {
...
+
+        require(FarmOwnerLike(farmOwner).farm() == farm, "FarmOwner farm mismatch");
+
+        FarmLike(farm).nominateNewOwner(farmOwner);
+        FarmOwnerLike(farmOwner).acceptOwnership();
+
+        dss.chainlog.setAddress(chainlogKey, farmOwner);
+    }
+
+    function initSBEBeam(
```
oldchili, line 302: Note that the Beam is not backward-compatible with other functions in this file, mainly `setFarm` (which changes the farm in the Splitter but doesn't update the Beam). I guess someone could claim it is a bug (though we usually treat the init functions as one-offs that are correct only for one use).

Maybe we should still add some comment like

```
Several functions in this file touch state a live SBEBeam depends on. They pre-date the Beam and should not be used as-is without considering it. For example, a spell that re-configures the module must account for the Beam alongside these calls:

• setFarm  — breaks set() (farm-sanity) until beam.file("farmOwner", …) re-points it.
• initSplitter/Kicker — temporarily overwrites Beam-managed hop/burn/kbump.
• kicker/splitter — immutable in the Beam; swapping either needs a new Beam.
```

oldchili: Maybe even a more general comment like:

"
Some functions in this file touch state a live Beam depends on. They pre-date the Beam and should not be used as-is without considering it. For example, `setFarm` does not consider the Beam.                                                                                                                                                         
                                                                                                                                                                                 
This is in-line with the regular init function methodology, where functions should be used for initial onboarding, then potentially be updated before a future use.
"

I think I like this one better.

sunbreak1211: Added

telome (dismissed):

oldchili (dismissed):
LGTM

src/SBEBeam.sol
```diff
@@ -0,0 +1,189 @@
...
+    // - It is intended to rewrite the same values, emit the event, and reset the toc count, even if there is no change.
+    // - Only the throughput-increasing directions are bounded: kbump is capped at maxKbump, hop is
+    //   floored at minHop, and the burn rate (kbump / hop) is capped at maxRate. Lowering kbump or
+    //   raising hop is otherwise allowed; at worst it stalls the burn stream, which governance can revive.
+    // - burn is capped at WAD (100%); a higher value would make Splitter.kick underflow and halt.
+    // - kbump must be a whole multiple of RAY, preserving the Kicker deploy invariant and avoiding kick dust.
+    // - hop is additionally capped below 5 years: a large enough hop makes the farm re-rate
+    //   rewardRate = leftover / hop truncate to 0. Then no further set() can revive those existing funds
+    //   (leftover = remaining * rewardRate). The cap also keeps hop well below type(uint256).max, the halt sentinel
+    //   (see the good modifier), so a facilitator cannot use set() to halt the engine and lock itself out.
```
oldchili:
```suggestion
    //   (see the set() require condition), so a facilitator cannot use set() to halt the engine and lock itself out.
```

src/SBEBeam.sol
```diff
@@ -0,0 +1,189 @@
...
+    // - Kicker.khump (the flap threshold) is deliberately left out of the set knobs: it is not a value that needs regular
+    //   tuning, and changing it is a more structural governance decision better routed through the full governance process.
+    function set(uint256 kbump, uint256 burn, uint256 hop) external toll {
+        require(splitter.live() == 1,                "SBEBeam/splliter-not-live");
```
oldchili:
```suggestion
        require(splitter.live() == 1,                "SBEBeam/splitter-not-live");
```

test/SBEBeam.t.sol
```diff
@@ -0,0 +1,583 @@
...
+    }
+
+    // --- set() gating ---
+
+    function testSetSplitterNotLive() public {
+        // A caged Splitter (live == 0) cannot execute kicks, so set() must not
+        // retune a dead engine; only governance can revive it.
+        vm.prank(pauseProxy);
+        splitter.cage(0);
+        vm.expectRevert("SBEBeam/splliter-not-live");
```
oldchili:
```suggestion
        vm.expectRevert("SBEBeam/splitter-not-live");
```

README.md
```diff
@@ -52,6 +52,39 @@ Configurable Parameters:
...
+
+`Splitter.burn` is bounded only at its upper end (`burn <= WAD`); it may be lowered freely down to zero.
+
+Note that `maxKbump` must be safe as a standalone kick size, not just safe relative to `minHop` (max rate). Because `hop` only gates the gap since the last kick (`Splitter.zzz`), a long-idle or freshly-deployed Splitter lets a kick fire immediately regardless of `hop`.
+
+Access control:
+* `wards` (`rely`/`deny`) - Governance-level administrators that configure the ranges.
+* `buds` (`kiss`/`diss`) - Facilitators permitted to call `set`.
+
+Halting: `set` is automatically blocked whenever the burn engine itself is stopped, i.e. when `Splitter.hop` is set to `type(uint256).max` (the canonical way governance disables kicks, since `Splitter.kick` then becomes unreachable). This binds the `SBEBeam` halt state to the engine's real halt state — there is no separate flag to keep in sync — and it ensures a facilitator can never use `set` to revive a governance-halted engine. Only governance can bring it back, by re-filing a finite `hop` on the `Splitter`.
```
oldchili:
```suggestion
Halting: `set` is automatically blocked whenever the burn engine itself is stopped, i.e. when `Splitter.hop` is set to `type(uint256).max` or `Splitter.live` is set to `0`. This binds the `SBEBeam` halt state to the engine's real halt state — there is no separate flag to keep in sync — and it ensures a facilitator can never use `set` to revive a governance-halted engine. Only governance can bring it back, by re-filing a finite `hop` on the `Splitter`.
```

oldchili (dismissed):

README.md
```diff
@@ -52,6 +52,39 @@ Configurable Parameters:
...
+
+`Splitter.burn` is bounded only at its upper end (`burn <= WAD`); it may be lowered freely down to zero.
+
+Note that `maxKbump` must be safe as a standalone kick size, not just safe relative to `minHop` (max rate). Because `hop` only gates the gap since the last kick (`Splitter.zzz`), a long-idle or freshly-deployed Splitter lets a kick fire immediately regardless of `hop`.
+
+Access control:
+* `wards` (`rely`/`deny`) - Governance-level administrators that configure the ranges.
+* `buds` (`kiss`/`diss`) - Facilitators permitted to call `set`.
+
+Halting: `set` is automatically blocked whenever the burn engine itself is stopped, i.e. when `Splitter.hop` is set to `type(uint256).max` or `Splitter.live` is set to `0`. This binds the `SBEBeam` halt state to the engine's real halt state — there is no separate flag to keep in sync — and it ensures a facilitator can never use `set` to revive a governance-halted engine. Only governance can bring it back, by re-filing a finite `hop` on the `Splitter`.
```
telome: Would maybe remove (or reword) the last sentence since it doesn't apply to `Splitter.live`.

```suggestion
Halting: `set` is automatically blocked whenever the burn engine itself is stopped, i.e. when `Splitter.hop` is set to `type(uint256).max` or `Splitter.live` is set to `0`. This binds the `SBEBeam` halt state to the engine's real halt state — there is no separate flag to keep in sync — and it ensures a facilitator can never use `set` to revive a governance-halted engine.
```

telome (dismissed):
Changes look good.

oldchili (approved):

telome (approved):

